PR #10, opened 2015-02-17
Minor docs fixes

lud4ik:
2,4) ...starting...  - Present Continious is used in form is/are + verb+"ing". I guess Present Simple is more appropriate.


tailhook:
Looks good. Thanks!

PR #15, opened 2015-03-10
Link librt and libutil as dynamic libs

rrader:
Using dynamic system libraries will make binary more portable, and on Arch linux there're only dynamic libraries of libc by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,8 @@ rust_compile_static = \
 	$(RUSTC) -o $(1).o --emit obj $(2); \
 	rlibs=$$($(RUSTC) -Z print-link-args $(2) \
 		| tr -s ' "' '\n' | grep rlib) \
-	&& gcc -static -static-libgcc $(1).o $$rlibs -o $(1) \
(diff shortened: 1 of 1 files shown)


tailhook:
Sorry, it doesn't work, because vagga is then run in absolutely empty container without those libraries. You should use "https://aur.archlinux.org/packages/glibc-static/" for building.

tailhook:
BTW, have you succeeded building for arch? It's possible to add Vagrantfile to the project so that you can build vagga in vagrant first time. Then you can rebuild vagga using vagga itself. Or perhaps I could provide a generic binary tarball in github releases page.

rrader:
I've submitted PKGBUILD for vagga: https://aur.archlinux.org/packages/vagga-git/
And kernel with CONFIG_USER_NS enabled https://aur.archlinux.org/pkgbase/linux-user-ns-enabled/

tailhook:
Nice. Could you also do a pull request to update vagga docs?

rrader:
Yep, added.
https://github.com/tailhook/vagga/pull/26

PR #18, opened 2015-03-31
Ubuntu non lts

simon-bourne:
Hey, here's the code so far for using non lts ubuntu. We still need to decide about what to do about the name/version both needing to be specified, but for now I'd appeciate any feedback on my rust code (it's the first time I've written any rust, so feel free to be critical)!

### DIFF
--- a/docs/build_commands.rst
+++ b/docs/build_commands.rst
@@ -239,9 +239,7 @@ work. To install a non LTS release, use:
 .. code-block:: yaml
 
     setup:
-    - !UbuntuRelease { name: utopic, version: 14.10 }
-
-The name and version number need to be consistent with one another.
+    - !UbuntuRelease { version: 14.10 }
 
 To install any ubuntu package use generic ``!Install`` command:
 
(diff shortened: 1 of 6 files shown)


tailhook:
From the first glance it lot ok. You don't need `concat!()`, it's possible to do wrap lines with slashes in current rust. (But I understand that my code was like so).

I take some time to think again about whether we want to require "name", because it's backward-compatibility issue.

simon-bourne:
Let me know what conclusions you come to. I found /etc/lsb-release which is in the cd images and is exactly what I want. I was only planning on looking at it when processing !UbuntuRelease, not !Ubuntu, so it shouldn't break anything out there already. The other thing to consider is do you want UbuntuRelease to look like this:

!UbuntuRelease { version: 14.10 }

or

!UbuntuRelease 14.10

I'm not too fussed which, but I'd slightly favour the first as it leaves the door open for future enhancments:

!UbuntuRelease { version: 14.10, kind: ... , arch: ... }

So you've got a "I just want an LTS Ubuntu" with !Ubuntu, and " I want to be more specific about ubuntu versions" with !UbuntuVersion, that can be extended.

tailhook:
Looks nice. Thanks! And yes, I like the extensible form.

simon-bourne:
Great, thanks!

PR #26, opened 2015-04-30
Docs for vagga installation on Arch Linux

rrader:
Installation of vagga using AUR packages is documented


tailhook:
Nice. But do you *always* need to add user to subuid/subgid? I think for users that created with new "shadow" package, subuid and subgid are filled automatically, just like in ubuntu. Am I right?

rrader:
Looks like you're right, will update doc

tailhook:
Merged. Thanks!

PR #42, opened 2015-07-09
WIP: add std::path::Path

joehillen:
I also alphabetized the std imports of affected files.

Errors when from 669 -> 321


joehillen:
Also, I've been targeting Rust 1.1.0. I assume that's what we want.

tailhook:
Sure 1.1 is ok. Merged it in. Thanks!

PR #44, opened 2015-07-10
Add Cargo.lock to repo

hauleth:


tailhook:
I'm not sure that adding Cargo.lock now is good idea, but we can update it any time anyway.

Also read_dir doesn't work like readdir worked. I.e. read_dir returns iterator over Result<Entry> values, but readdir returned list of Path values. I.e. you need to try!() on each iteration, and presumably resolve entry.path()

Anyway I'm merging it in, to keep process moving :)

PR #48, opened 2015-08-18
Fix some warnings

joehillen:
Mostly unused imports and unused mut.

@tailhook This one looks like a bug https://github.com/tailhook/vagga/blob/master/src/container/async.rs#L82 

```
src/container/async.rs:81:35: 81:43 warning: unused variable: `duration`, #[warn(unused_variables)] on by default
src/container/async.rs:81     pub fn add_timeout(&mut self, duration: Duration, name: Name) {
```

### DIFF
--- a/src/builder/capsule.rs
+++ b/src/builder/capsule.rs
@@ -9,7 +9,7 @@ use std::collections::HashSet;
 use std::fs::{File};
 use std::io::{Write};
 use std::path::Path;
-use std::process::{Command, ExitStatus, Stdio};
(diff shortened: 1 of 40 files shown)


tailhook:
Yes. It's definitely a bug. It works because the method is only ever used with zero timeout :)

Merged. Thanks!

joehillen:
I opened an issue for it so that it doesn't get forgotten. #49 

I'm not sure how duration should be added there, so I'll let you fix it.

PR #50, opened 2015-08-19
Archlinux AUR bin package docs

rrader:
Documentation for Arch Linux installation steps is fixed to guide user to use binary AUR package instead of broken git.


tailhook:
Thanks!

PR #51, opened 2015-08-19
Fix prefix variable in install.sh script

rrader:
Now prefix equals to DESTDIR which brakes the build.


tailhook:
Nice catch. Thanks!

PR #109, opened 2015-10-30
Example with PostgreSQL+Alembic added

rrader:
Using the case of application using alembic migrations, this example shows how to perform synchronization between containers using filesystem.

Here, container with app waits until database sets up and runs migration after that.

Not sure if there's neater way to do it.


tailhook:
Looks fine. Thanks!

tailhook:
Well, It's also good idea to add a little description into `docs/examples`. So that it's visible in http://vagga.readthedocs.org/en/latest/examples.html This may optionally include tips or commands to run. @rrader, would you like to do that?

PR #129, opened 2015-11-27
temporarily remove another non-working alpine mirror

popravich:


tailhook:
The problem with alpine mirrors is ugly :( But thanks anyway!

Issue #126, opened 2015-11-25
Running GUI applications in vagga


gabrielelanaro:
Hi, I would like to know if it is possible and what are the steps to run GUI applications (eg Firefox) in vagga. The idea would be to share the video device and screen with the host system (I'd like to test/develop webgl and opengl applications)

tailhook:
Hi,

Well, you should be able to run GUI applications. There are few issues however

(I) For the X.org socket, you have two options:

1. Your X11 socket is in abstract network namespace. It is in some distributions. And this should make GUI applications work as long as you run `vagga --use-env DISPLAY <command>`
2. You add `/tmp/.X11-unix/..` socket into `external-volumes` setting and mount it inside the container.

(2) Shared memory is separate in a container. Usually, it's okay, but it's probably important for OpenGL. I'm not sure if we can figure out a way to keep shared memory shared with host system

(3) I'm not sure about access to video device(s). It should inherit same capabilities as your user host system, but it looks like that sometimes doesn't work. For example `fuse` doesn't work in user namespaces at all and it can't be fixed AFAIK

(4) Similar to (1) but about dbus socket if you need (I'm not sure about firefox)

Overall GUI applications should work. But I've never tried it myself. And it will probably require some effort to setup this now and to reintegrate some enhancements to the vagga to get this working easily. Would be nice, if you try and give some feedback on this.

gabrielelanaro:
Thanks so much for your answer! I'm trying to setup the GUI integration to know at least how far we can get.

Right now I'm having some problems with creating/mounting the .X11-unix (I'm using the testing automated installer)

This is the one I'm using:
```
containers:
  sphinx:
    setup:
    - !Ubuntu trusty
    - !EnsureDir /tmp/.X11-unix/
#    - !Install [firefox]
#    volumes:
#      /tmp/.X11-unix: !BindRW /volumes/X11

commands:
  bash: !Command
    container: sphinx
    run: /bin/bash
```

If I do vagga bash, I get to a shell but /tmp/.X11-unix is not present. But if I do find the directory in another location:
```
root@gabriele-desktop:/work/# find . -name .X11-unix
./.vagga/.roots/sphinx.4ac63aa4/root/tmp/.X11-unix
```

tailhook:
You can't `EnsureDir` in `/tmp` because the `/tmp` is mounted as tmpfs filesystem on container start. You can use builtin feature of `Tmpfs` by overriding `/tmp` mount:
```
containers:
[...]

PR #131, opened 2015-11-30
Examples for firefox #126

gabrielelanaro:
There are three examples. Simple firefox, and installation of graphics driver for intel and nvidia cards. (issue #126 )

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -29,6 +29,7 @@ Miscellaneous
 
    examples/misc/travis
    examples/misc/selenium
+   examples/misc/firefox
 
 
 Documentation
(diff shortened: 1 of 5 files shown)


tailhook:
Thanks! Couple of tweaks:

1. `HOME` should be set in environment of container/command, it's useless on command-line
2. Maybe there is an "install most drivers" meta package on Ubuntu? (I'm not the ubuntu user, so I'm not sure) I.e. the one that will work for most users. Usually, the whole point of vagga is to be able to run a project on any computer, rather than everybody fixing vagga.yaml on its own. What do you think?

gabrielelanaro:
1) I updated the configuration in the yaml files
2) for my specific video card I tried to install the generic drivers (for example nouveau and the nvidia present in repository) but that didn't work, I don't know if there exist a universal recipe that would work on any computer, the problem is that this kind of configuration is truly hardware dependent, it just won't work on any machine anyway.

tailhook:
Fine. Thanks for taking the time!

PR #137, opened 2015-12-07
Providing some usage examples of undocumented commands in Build Steps

lilianmoraru:
I guess, some notes/comments should be removed when detailed descriptions of the commands are written.
Some commands probably need longer code examples, so I left them(ex: CacheDirs, Container, SubConfig). I was not sure even if these are actually ok, if they are not to long...


tailhook:
I think they look pretty well. Thanks!

PR #145, opened 2015-12-16
Docker2vagga script improvements

anti-social:


tailhook:
You should run `vagga test tests/subconfig.bats` (at least) to check it. It looks like that merely a container's checksum is changed, still I can't apply the pull request as is.

PR #148, opened 2015-12-21
Changed latest alpine version to 'v3.3', added test for Alpine 3.3

anti-social:


tailhook:
This is probably fine, but we usually update `./fetch_binaries.sh` too with the new release of alpine. Because in this case you fetch new apline packages with the old `apk` tool.

Issue #127, opened 2015-11-25
Warn about EnsureDir in one of the volumes


tailhook:
Inspired by #126

PR #150, opened 2015-12-22
#127 Warn about EnsureDir in one of the volumes

anti-social:


tailhook:
Looks fine, except `format("'{0}'", path.display())` formatting. That was useful in old rust. You should use `format("{:?}", path)` in new rust, it represents paths well enough, and adds quotes on it's own.

tailhook:
Merged in 18b6341

Issue #152, opened 2015-12-28
Ctrl+C should not kill processes which look like a shell (like python console or bash)


tailhook:
It looks like some bug of how we handle sessions and controlling terminal. Probably some thing should be implemented in `unshare`

tailhook:
@anti-social seems to work on this issue

PR #162, opened 2016-01-26
 #152 Handle Ctrl+C for _run and user commands

anti-social:


anti-social:
Also we should correctly handle Ctrl+Z, but there some troubles with it. Vagga wrapper can't send SIGSTOP to self. So wrapper should ask parent process that is outside of namespace to stop wrapper.

src/tty_util.rs
```diff
@@ -0,0 +1,30 @@
...
+pub fn give_tty_to(fd: c_int, pid: pid_t) -> Result<(), String> {
+    let res = unsafe { ioctl(fd, TIOCSPGRP, &pid) };
+    match res {
+        res if res < 0 => Err(format!("Error when giving tty with fd {} to process {}", fd, pid)),
+        _ => Ok(()),
+    }
+}
+
+pub fn ignore_tty_signals() -> Result<(), String> {
+    let tty_signals = vec![SIGTTIN, SIGTTOU];
```
tailhook: It's not "rusty" to allocate a vector just to iterate over the items. Should just call `sigaction()` twice. (And yes `sigaction` is a safe wrapper around the same stuff.

src/wrapper/mod.rs
```diff
@@ -80,6 +83,34 @@ pub fn run() -> i32 {
         }
     };
 
+    let tty_fd = STDIN_FILENO;
+    let is_interactive = isatty(tty_fd).unwrap_or(false);
+    if is_interactive {
```
tailhook: All this stuff should be wrapped into a function. And I believe it's better to use that function directly in _run, _run_in_netns, commandline_cmd.

It's good idea to mention in commit message that supervise commands are not implemented yet. (I think it's doable right?)

src/process_util.rs
```diff
@@ -97,6 +101,66 @@ pub fn run_and_wait(cmd: &mut Command)
...
+        Ok(child) => child,
+        Err(e) => return Err(format!("Error running {:?}: {}", cmd, e)),
+    };
+    let mut trap = Trap::trap(&[SIGCHLD, SIGTERM]);
+    'signal_loop: for signal in trap.by_ref() {
+        match signal {
+            SIGCHLD => {
+                debug!("Received SIGCHLD signal, waiting child");
+                loop {
+                    match waitpid(child.pid(), Some(WNOHANG | WUNTRACED)) {
```
tailhook: You wait only on single child. But other children may die too (even if we have not created them). This code pobably leaves zombie processes behind.

src/process_util.rs
```diff
@@ -97,6 +101,66 @@ pub fn run_and_wait(cmd: &mut Command)
...
+                        Ok(WaitStatus::Stopped(_, signum)) => {
+                            if signum == SIGTTOU || signum == SIGTTIN {
+                                try!(give_tty_to(tty_fd, child.pid()));
+                                try!(child.signal(SIGCONT)
+                                     .map_err(|e| format!(
+                                         "Error when sending SIGCONT
+                                          signal to {:?}: {}", cmd, e)));
+                            }
+                            continue 'signal_loop;
```
tailhook: SIGCHLD is old-style signal, which means if it raised multiple times it is signalled only once. Then it's expected that you will waitpid() all the children before waiting for the next signal.

src/tty_util.rs
anti-social: Can't use sigaction from nix package together with SIG_IGN: https://github.com/carllerche/nix-rust/pull/241

src/wrapper/mod.rs
anti-social: We should always make wrapper a group leader otherwise some tests is hung. The reason is that `vagga _build` that is run after `vagga _run` will be stopped when accessing stdin due to lack of controlling tty.

anti-social:
Should use `killpg` to send `SIGCONT` signal. TODO: fix and add test case.

src/process_util.rs
```diff
@@ -64,31 +64,57 @@ pub fn capture_fd3_status(mut cmd: Command)
...
+                try!(child.signal(SIGINT)
+                     .map_err(|e| format!(
+                         "Error sending SIGINT to {:?}: {}", cmd, e)));
             }
             SIGTERM => {
                 // SIGTERM is usually sent to a specific process so we
                 // forward it to children
                 debug!("Received SIGTERM signal, propagating");
-                child.signal(SIGTERM).ok();
+                try!(child.signal(SIGTERM)
```
tailhook: Why do you think it's a fatal error? The error was ignored because we can't ensure that process is still alive.

tailhook:
Merged as part of #168

PR #168, opened 2016-02-12
Interrupt handling

### LINKED ISSUE #152: Ctrl+C should not kill processes which look like a shell (like python console or bash)

tailhook:
Presumably, this supersedes #162 and fixes #152, changes are:

1. Add `child_events` handler to `unshare` which is similar to `reap_zombies` but also allows to trap stops. It looks much more clear than handling children in TtyGuard
2. Always use `make_group_leader` when starting process where needed, instead of making itself a leader in TtyGuard
3. TtyGuard knows whether stdio is a tty by probing, without external setup, and it's transparent for the user
4. TtyGuard also detects whether one of stdout or stderr is a tty, even if stdin is not

Other things changed:
1. SIGTTIN and SIGTTOU are masked out rather than ignored. It's a matter of preference, though (I just use less code)
2. All commands which may consume input run from vagga build must be run with `run_and_wait`. Skipping it, unfortunately, results into hanging processes (and I have not checked all the places)
3. Processes become group leaders even when run not on the tty. Shouldn't be a problem, right?

Probably we need to use /dev/null for stdin in build process so that (2) from above doesn't get triggered.

Some more polishing might be needed:

*  Vagga is less responsive to Ctrl+C with this patch. Probably a signal is handled by the application and vagga continues to work after application has shut down. Perhaps previously vagga sometimes just dropped the container without proper waiting for exit.
* I'm also not sure that we need to propagate a signal by hand to the children processes.


anti-social:
Now Ctrl+C is sent to vagga, not child process. So for example ipython can't handle SIGINT.

anti-social:
We should also send SIGINT signal to the child process group.

tailhook:
> Now Ctrl+C is sent to vagga, not child process. So for example ipython can't handle SIGINT.

This scenario is tested and worked for me

> We should also send SIGINT signal to the child process group.

Ah, well, fair enough. But the code to do that is still in the pull request.

Issue #176, opened 2016-02-17
Completion for options should work only when dash is written


tailhook:
Similarly to how we complete built-in commands only on underscore.

Typing single (and obvious) char is not a problem, but searching for names of few commands in large list of options is ugly.

/cc @anti-social

lilianmoraru:
@tailhook This can be closed?

tailhook:
Yes.

PR #177, opened 2016-02-17
#176 Autocomplete options only if dash is written

anti-social:


tailhook:
Nice. Thanks!

lilianmoraru:
`install: cannot stat ‘vagga-completion.sh’: No such file or directory` - when installing `vagga-testing`

PR #185, opened 2016-03-03
PHP/Composer support

livioribeiro:
This pull request adds support for PHP and the Composer package manager.


tailhook:
Nice work!

The only issue I have is this:

> This command will install packages (including dev) from ``composer.json`` into
 +``/work/vendor`` using Composer's ``install`` command

The problem that it interfere with how containers are rebuilt and versioned. As a thought experiment, what happens if you have two containers, with different dependencies, what will be in the "vendor" folder.

We have a similar issue with npm. And npm has `--global` flag for installation in the system. Which vagga always uses. Sometimes this means we need to write `NODE_PATH` in environment or `/usr/local/lib/node_modules` in `webpack.config.js` but this is a small price to pay for good versioning.

livioribeiro:
The tests failed when checking the hash of the containers running HHVM ('[[ $link = ".roots/hhvm-ubuntu-trusty.76604d51/root" ]]' failed). I don't know why but they seem unpredictable.

> Nice work!
> 
> The only issue I have is this:
> 
> >    This command will install packages (including dev) from composer.json into
> >    +/work/vendor using Composer's install command
> 
> The problem that it interfere with how containers are rebuilt and versioned. As a thought experiment, what happens if you have two containers, with different dependencies, what will be in the "vendor" folder.
> 
> We have a similar issue with npm. And npm has --global flag for installation in the system. Which vagga always uses. Sometimes this means we need to write NODE_PATH in environment or /usr/local/lib/node_modules in webpack.config.js but this is a small price to pay for good versioning.

That makes sense. I did this way because PHP is not nice when it comes to requiring libs from a location other the current dir, requiring the 'include_path' to be set either in the config file (php.ini) or in application code (there is no PHP_PATH afaik), and also because a lot of php deployments are still done in shared hosts.

I'll try to edit the config file automatically during build, the worst case would be to warn the users to set the 'include_path'.

tailhook:
Hm, what about putting `/etc/php.d/vagga.ini` ? It looks like php should use the config by default

livioribeiro:
Yes, you're right, I forgot about that!

tailhook:
> The tests failed when checking the hash of the containers running HHVM ('[[ $link = ".roots/hhvm-ubuntu-trusty.76604d51/root" ]]' failed). I don't know why but they seem unpredictable.

This is because you use `composer.lock` which is generated during the build. The command should either use (and require) a `composer.lock` or use `composer.json` (and somehow make composer ignore a `composer.lock`). I'm not sure there is a way to version `composer.lock` which is generated during the build in any predictable way.

livioribeiro:
But all HHVM tests are failing, even those that are not using composer.lock

livioribeiro:
I've made some changes:
* Dependencies are installed globally
* Binaries installed by composer are automatically symlinked into PATH
* Composer itself is symlinked into PATH and removed after container is built
* `include_path` is automatically configured
* Added option to override `include_path`

tailhook:
Sounds great. Any issues with hash sums left?

livioribeiro:
I left just 1 test using hhvm (the only tests with hash problems were hhvm), they didn't make a lot of sense since it is not supported neither by the pull request nor by any distro, and the test that remained seems to be ok

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+            .map(|l| l.to_owned())
+            .collect();
+
+        if !conf_d_lines.is_empty() {
+            conf_d_lines
+        } else {
+            let args = [
+                "/bin/sh".to_owned(),
+                "-exc".to_owned(),
+                "php --ini | grep -E 'conf.d$' | cut -d ':' -f 2 | cut -d ' ' -f 2".to_owned()
```
tailhook: We don't do shell pipelines in vagga. We assume that there might be no shell in container... Just capture and parse it by rust. It wouldn't be too hard, I swear :)

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    Ok(())
+}
+
+fn setup_include_path(ctx: &mut Context) -> Result<(), String> {
+    let args = [
+        "/bin/sh".to_owned(),
+        "-exc".to_owned(),
+        "find $(ls -d /etc/php*) -name 'conf.d' | grep -E '^/etc/php'".to_owned(),
```
tailhook: Well, I'm not even sure what it does. Why is it so complex?

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+    ];
+    try!(run_command(ctx, &args));
+
+    let args = [
+        "ln".to_owned(),
+        "-s".to_owned(),
+        "/tmp/composer.phar".to_owned(),
+        "/usr/local/bin/composer".to_owned(),
+    ];
+    try!(run_command(ctx, &args));
```
tailhook: this should be `std::os::unix::fs::symlink`

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+    try!(copy(&composer_inst, &Path::new("/vagga/root/tmp/composer-setup.php"))
+        .map_err(|e| format!("Error copying composer installer: {}", e)));
+
+    let runtime_exe = ctx.composer_settings
+        .runtime_exe
+        .clone()
+        .unwrap_or(DEFAULT_RUNTIME.to_owned());
+
+    let args = [
+        "/usr/bin/env".to_owned(),
```
tailhook: Why do you need `/usr/bin/env` ?

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    let include_path = ctx.composer_settings.include_path
+        .clone()
+        .unwrap_or(DEFAULT_INCLUDE_PATH.to_owned());
+
+    for conf in conf_d {
+        let args = [
+            "/bin/sh".to_owned(),
+            "-exc".to_owned(),
+            format!("echo 'include_path={}' > {}/vagga.ini", include_path, &conf),
```
tailhook: same thing, just create a file

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    Ok(())
+}
+
+pub fn finish(ctx: &mut Context) -> Result<(), StepError> {
+    try!(list_packages(ctx));
+    try!(run_command(ctx, &[
+        "rm".to_owned(),
+        "/usr/local/bin/composer".to_owned(),
+    ]));
```
tailhook: And just remove a file

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,252 @@
...
+            ];
+            let conf = try!(capture_command(ctx, &args, &[])
+                .and_then(|result| {
+                    String::from_utf8(result)
+                    .map_err(|e| format!("{}", e))
+                }));
+            let args = [
+                "mkdir".to_owned(),
+                "-p".to_owned(),
+                conf.clone(),
```
tailhook: Same thing with directories. There is a function `file_util::create_dir`.

src/builder/commands/composer.rs
livioribeiro: Sorry about that, I'll fix it asap

src/builder/commands/composer.rs
livioribeiro: In ubuntu, there is a conf.d for cli, cgi, apache, etc. and I was not sure if it would accept just /etc/php5/conf.d. I'll figure out a better solution

src/builder/commands/composer.rs
livioribeiro: In my tests ubuntu didn't find `php` command, it has to `php5` so I thought it would be safer to use `/usr/bin/env`. I'll take a look at that again to see if it is really needed

livioribeiro:
The issues were fixed, I hope everything's ok now.

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,269 @@
...
+        .unwrap_or(DEFAULT_RUNTIME.to_owned());
+
+    let args = [runtime_exe, "--ini".to_owned()];
+    let output = try!(capture_command(ctx, &args, &[])
+        .and_then(|x| String::from_utf8(x)
+            .map_err(|e| format!("Error parsing command output: {}", e)))
+        .map_err(|e| format!("Error reading command output: {}", e)));
+
+    // match any line that ends with /etc/php*/**/conf.d, get first result
+    let re = try_msg!(Regex::new(r#"(?m).*?(/etc/php\d/.*?conf.d)$"#), "Invalid regex: {err}");
```
tailhook: this should use `.expect("invalid regex")` rather than `try_msg`. try_msg is for runtime errors. But regex compilation will fail only when you have put wrong regex (which will be caught by unit tests)

src/builder/commands/composer.rs
```diff
@@ -0,0 +1,269 @@
...
+    Ok(())
+}
+
+fn create_vagga_ini(location: &Path, content: &str) -> Result<(), String> {
+    File::create(location)
+        .and_then(|mut f| f.write_all(content.as_bytes()))
+        .map_err(|e| format!("Error creating file {:?}: {}", location, e))
+}
+
+fn find_dirs(ctx: &mut Context, search: &str) -> Result<Vec<PathBuf>, String> {
```
tailhook: I'm concerned about this implementation.

Why can't you list directory `/etc` and find all `php.*` in there?

Anyway, there is a `scan_dir` crate which is already linked into vagga. It has both recursive and non-recursive directory traversal. I believe non-recursive should be enough, though.

tailhook:
Okay, except find_dirs, it's fine. I will take a closer look to your versioning code later (perhaps this night).

src/builder/commands/composer.rs
livioribeiro: In Ubuntu, there is a `conf.d`s for each environment, like `/etc/php/cli/conf.d` for cli, `/etc/php/cgi/conf.d` for cgi, `/etc/php/apache2/conf.d` for apache, but there's no global `conf.d`.

`scan_dir` with recursive traversal will do the job.

tailhook: Okay. Why recursive? Just walk through `/etc/php/*` and look for `conf.d` in each of them.

livioribeiro:
> What if this is /vagga/root/etc/python/conf.d ?
> 
> By the way, if you need to visit two levels of hierarchy you don't need recursive traversal too. Just put two nested loops. like this:

Oh, sorry, I `git commit`ed a mistake, it was supposed to check the php directory manually and then use it to find the `conf.d`s inside it.

But that code was ugly, so I followed your example and came up with this (which also tries to find conf.d at the root of /etc/php):

```rust
ScanDir::dirs().read("/vagga/root/etc", |iter| {
    iter.filter(|&(_, ref name)| name.starts_with("php"))
    .flat_map(|(ref entry, _)| {
        ScanDir::dirs().read(entry.path(), |iter| {
            iter.filter(|&(ref entry, ref name)| {
                name == "conf.d" ||
                entry.path().join("conf.d").exists()
            })
            .map(|(ref entry, _)| {
                let path = entry.path();
                if path.ends_with("conf.d") { path }
                else { path.join("conf.d") }
            })
            .collect()
        })
    })
    .collect()
})
```

But now I got a problem with Ubuntu, the `ScanDir::dirs().read("/vagga/root/etc", |iter| {` is failing with the message `error reading directory "/vagga/root/etc/nologin": No such file or directory (os error 2)`.

It works fine with Alpine, though.

tailhook:
Perhaps it's broken symlinks. Adding `dirs().skip_symlinks(true)` would probably be good enough. Properly handling of symlinks will require `temporary_change_root` function.

tests/composer.bats
```diff
@@ -0,0 +1,114 @@
+setup() {
+    cd /work/tests/composer
+}
+
+teardown() {
+    cd /work/tests/composer
+    if [ -d vendor ]; then rm -r vendor; fi
+    if [ -f composer.lock ]; then rm composer.lock; fi
+    cd /work/tests/composer_lock
+    if [ -d vendor ]; then rm -r vendor; fi
```
tailhook: Is all of this still needed? I thought you install the packages to the system place

tailhook:
We also probably need to encapsulate versioning of the command to a function. That should fix crash in tests (yes it looks like unrelated but it is here because stack space is reserved even when that branch of match is not visited in the specific test).

I'm not sure that versioning `composer.lock` is a good idea, but I'm not sure of the opposite too.

These two things, and I'm going to squash and merge it in.

tests/composer.bats
livioribeiro: I forgot to remove those lines, only the line that removes the lock file is needed

livioribeiro:
> We also probably need to encapsulate versioning of the command to a function

Just a function on the same module or should I create a separate module?

> I'm not sure that versioning composer.lock is a good idea, but I'm not sure of the opposite too.

I believe it is needed because versions in the lock file can be updated without changing the main file, and if someone pushes an updated `composer.lock` to vcs, the container will need to be rebuilt to update dependencies.

And I just learned that the hashes present in the lock file weren't what I thought they were, those are hashes of the `composer.json`, so I will have to do the hash of the lock file.

tailhook:
> Just a function on the same module or should I create a separate module?

Any option will fix the issue. Creating a module is probably better. 

> I believe it is needed because versions in the lock file can be updated without changing the main file, and if someone pushes an updated composer.lock to vcs, the container will need to be rebuilt to update dependencies.

Okay, let's try this way.

tailhook:
Merged. Thanks!

PR #192, opened 2016-03-11
Added support for different Ubuntu architectures.

jpiles:
I'm a total Rust newbie, so I hope I've used the proper idioms. 

I've tested it with UbuntuRelease (with and without the "arch" option) and with plain "Ubuntu", and it seems to work as expected.

I haven't found how to build the docs, though, but I guess it'll be OK.

### DIFF
--- a/docs/build_steps.rst
+++ b/docs/build_steps.rst
@@ -63,6 +63,9 @@ Ubuntu Commands
      The verison of ubuntu to install. This must be digital ``YY.MM`` form,
      not a code name. **Required**.
 
+   arch
+     The architecture to install. Defaults to ``amd64``.
+
    keep-chfn-command
      (default ``false``) This may be set to ``true`` to enable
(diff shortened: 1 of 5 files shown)


tailhook:
The pull request is okay, but please add a test. One test would be good enough. Tests are in `test/ubuntu.bats` and `tests/ubuntu/vagga.yaml`, to run just ubuntu tests use `vagga test tests/ubuntu.bats`

> I haven't found how to build the docs, though, but I guess it'll be OK.

Your changes are okay. But you don't have to remember how to do it, that's the whole point of vagga :)
Just try:
```console
$ vagga
 Available commands:
...
    doc                 Build vagga documentation
$ vagga doc
...
Documentation is built under docs/_build/html/index.html
```

jpiles:
Thanks for the doc tip! I was trying vagga make doc or something :)

tailhook:
Squashed and merged in 08dc3d6

PR #195, opened 2016-03-11
_push_image command

anti-social:


vagga.yaml
```diff
@@ -148,6 +148,7 @@ commands:
       if [ -n "$ALPINE_MIRROR" ]; then
           echo "alpine-mirror: $ALPINE_MIRROR" >> $HOME/.vagga.yaml
       fi
+      echo "push-image-cmd: curl -T \${source_path} http://localhost:8000/images/\${container_name}.\${hash}.tar.xz" >> $HOME/.vagga.yaml
```
tailhook: could use single quotes instead of escaping

tailhook:
Well, what's the downsides of using stderr for `capsule` always? The upside is obviously smaller number of unuseful arguments.

src/main.rs
```diff
@@ -47,6 +48,7 @@ fn main() {
         Some("vagga_version") => version::main(),
         Some("vagga_wrapper") => wrapper::main(),
         Some("vagga_build") => builder::main(),
+        Some("vagga_pack") => pack::main(),
```
tailhook: I don't think we should add more variants here. Is there any lightweight alternatives? Do you actually need to spawn a process to perform the task?

This was just a legacy when migrating from vagga 0.2  (which contained multiple binaries) to vagga 0.3 (which has single binary for everything)

src/options/build_mode.rs
```diff
@@ -19,6 +20,9 @@ impl Default for BuildMode {
 pub fn build_mode<'x>(ap: &mut ArgumentParser<'x>, mode: &'x mut BuildMode)
 {
     ap.refer(mode)
+    .add_option(&["--no-image"], StoreConst(BuildMode::NoImage), "
```
tailhook: Name is misleading. `--no-external-cache`, or `--skip-cache-url` or `--ignore-cache` or `--no-cache-check`, `--no-image-download`

We want to stress that it is a "cache" and it's external/network/download

src/options/pack.rs
```diff
@@ -32,6 +34,9 @@ impl Options {
             ap.refer(&mut opt.file)
                 .add_option(&["-f", "--file"], ParseOption,
                     "File to store tar archive at");
+            ap.refer(&mut opt.compression_type)
+                .add_option(&["-t", "--compression-type"], ParseOption,
```
tailhook: Probably the short argument `-t`. If we will ever need a short argument for this option, I would prefer `tar`-compatible options `-zJj`

src/pack/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+            ");
+        ap.refer(&mut container)
+          .add_argument("container", Store,
+                "A container to pack")
+          .required();
+        ap.refer(&mut file_path)
+            .add_option(&["-f", "--file"], ParseOption,
+                "File to store tar archive at");
+        ap.refer(&mut compression_type)
+            .add_option(&["-t", "--compression-type"], ParseOption,
```
tailhook: This is similar. Remove `-t`

src/wrapper/build.rs
```diff
@@ -292,6 +301,9 @@ pub fn build_container_cmd(wrapper: &Wrapper, cmdline: Vec<String>)
         ap.refer(&mut force)
             .add_option(&["--force"], StoreTrue,
                 "Force build even if container is considered up to date");
+        ap.refer(&mut no_image)
+            .add_option(&["--no-image"], StoreTrue,
```
tailhook: Similarly to the above. In this case simpler `--no-download` may be good enough.

src/config/read_settings.rs
```diff
@@ -20,6 +20,7 @@ struct SecureSettings {
     alpine_mirror: Option<String>,
     site_settings: BTreeMap<PathBuf, SecureSettings>,
     external_volumes: HashMap<String, PathBuf>,
+    push_image_cmd: Option<String>,
```
tailhook: Well, it believe it should be `push-image-script`. Because (a) it's not abbreviated and (b) in current implementation it's more like script rather than command: a complex expression with substitutions is expected (or a path to external script which should work as well), and not like a command like `rsync` which has well-known arguments which we can add arbitrarily.

src/launcher/push.rs
```diff
@@ -0,0 +1,87 @@
...
+    };
+    let source_path = PathBuf::from(".vagga")
+        .join(&roots)
+        .join(&ver)
+        .join("image.tar.xz");
+    match settings.push_image_cmd {
+        Some(ref push_image_cmd) => {
+            let mut upload_cmd = Command::new("/bin/sh");
+            upload_cmd.stdin(Stdio::null())
+                .arg("-c")
```
tailhook: It should be at least `-ec` as it helps with common errors so much, that we use it in vagga almost always. Perhaps it's a good idea to use `-exc` as in many other places. This helps users to debug their scripts.

anti-social:
> Well, what's the downsides of using stderr for capsule always? The upside is obviously smaller number of unuseful arguments.

If you run `vagga _build` it will be strange to see some output in stdout and other in stderr. So maybe it is worth to use stderr for any `_build` output?

tailhook:
Well, I'm not so sure about all build output. Especially if still allow to read from stdin (i.e. have interactive builds).

anti-social:
> I don't think we should add more variants here. Is there any lightweight alternatives? Do you actually need to spawn a process to perform the task?

There was a problem when I needed `curl` inside `capsule`. There were some permission errors when installing `alpine-baselayout`:

```
ERROR: Failed to set ownership on etc/shadow.apk-new: Invalid argument
Executing alpine-baselayout-2.3.2-r10.post-install
chown: /etc/shadow: Invalid argument
Executing busybox-1.24.1-r7.trigger
1 errors; 1 MiB in 5 packages
ERROR:vagga::wrapper: Error executing _pack_image: Apk exited with status: exited with code 1
```

I think it's because `capsule` doesn't set uidmap for commands.

tailhook:
AFAIU, it does set uidmap, but different than one in container. I.e. it maps uids one to one to external uids, so usually there is a gap 1..100000 in the uids, which normal OS use. Well, maybe we should rethink it, but it's non-trivial task. Can you make curl work without alpine-baselayout?

anti-social:
Now we don't need `curl` inside `capsule` cause we use `push-image-script` setting. I've just checked that `capsule` works fine when installing gzip, bzip2, xz.

anti-social:
`apk_run` is used from alpine builder. So I'm not sure that we should always redirect `apk_run` output to stderr.

tailhook:
> apk_run is used from alpine builder.

Sure. It's not a problem. Sometimes I even think we should hide it from users at all :)

Anyway, I'm leaning towards disable stdin, and redirect whole build to stderr. Because `vagga command > file` should put command output into the file, not artifacts of the build process.

Also, with current management of process groups we gen uninterruptable builds sometimes. Which is wrong. Disabling stdin and group management during build should fix the issue.

tailhook:
Okay, looks fine, except these things:

1. I still don't like `image-url`. Let's call it `image-cache-url`
2. I think we should redirect stdout to stderr in `vagga_build` before applying this patch (I can't figure out current semantics easily, and I don't like such intrusive use of RawFd)

tailhook:
Okay, just committed ``process_util::squash_stdio`` which is used for build and should be used for `_pull_image` too.

src/builder/capsule.rs
```diff
@@ -37,8 +42,10 @@ pub struct State {
 
 // Also used in alpine
 pub fn apk_run(args: &[&str], packages: &[String]) -> Result<(), String> {
+    let stdout_fd = try!(dup(2).map_err(|e| format!("{}", e)));
     let mut cmd = Command::new("/vagga/bin/apk");
     cmd.stdin(Stdio::null())
+        .stdout(unsafe { Stdio::from_raw_fd(stdout_fd) })
```
tailhook: squash_stdio here too

src/wrapper/pack.rs
```diff
@@ -22,24 +24,48 @@ pub fn pack_image_cmd(wrapper: &Wrapper, cmdline: Vec<String>)
     let cconfig = try!(wrapper.config.containers.get(&options.name)
         .ok_or(format!("Container {} not found", options.name)));
     let container_ver = wrapper.root.as_ref().unwrap();
-    let root = Path::new("/vagga/base/.roots")
-        .join(container_ver).join("root");
 
-    let mut cmd = Command::new("/vagga/bin/busybox");
-    cmd.stdin(Stdio::null())
+    let tmppath = PathBuf::from(
+        &format!("/vagga/base/.roots/.tmp.{}", &options.name));
```
tailhook: I'm not sure about this name. The name is reserved for building containers. Can we just put it to the tmpfs?

src/wrapper/pack.rs
```diff
@@ -22,24 +24,48 @@ pub fn pack_image_cmd(wrapper: &Wrapper, cmdline: Vec<String>)
...
+        } else {
+            tar_cmd.arg(Path::new("/work").join(f));
+        }
+    }
+    tar_cmd
+        .arg("-C").arg(&root)
+        .arg(".");
+
+    if capsule_features.len() > 0 {
+        let mut ctx = BuilderContext::new(&wrapper.config, options.name.clone(),
```
tailhook: Looks like huge hack. I think we could break down `ensure_features` and pass just `capsule::State` and settings instead of creating a build context.

src/builder/mod.rs
```diff
@@ -69,48 +79,95 @@ pub fn run() -> i32 {
...
+        if let Some(ref image_cache_url_tmpl) = container.image_cache_url {
+            let hash = match ver.rsplitn(2, ".").next() {
+                Some(v) => v,
+                None => {
+                    error!("Incorrect container version");
+                    return 122;
+                }
+            };
+            let image_cache_url = image_cache_url_tmpl
+                .replace("${container_name}", &container_name)
```
tailhook: Why not just braces?

src/launcher/push.rs
```diff
@@ -0,0 +1,87 @@
...
+                    return Err(format!("Error when uploading image: {:?}", upload_cmd));
+                },
+                Err(e) => {
+                    return Err(format!("Error when uploading image: {}", e));
+                },
+                _ => {},
+            }
+        },
+        None => {
+            return Err(format!("You should specify 'push-image-cmd' setting"));
```
tailhook: Should fix the error message

src/launcher/push.rs
```diff
@@ -0,0 +1,87 @@
...
+        .join(&roots)
+        .join(&ver)
+        .join("image.tar.xz");
+    match settings.push_image_script {
+        Some(ref push_image_script) => {
+            let mut upload_cmd = Command::new("/bin/sh");
+            upload_cmd.stdin(Stdio::null())
+                .arg("-exc")
+                .arg(push_image_script)
+                .env("source_path", source_path.to_str().unwrap())
```
tailhook: The `source_path` is the bad name. This looks like it relates to source code somehow. Maybe `tmp_image_path`

tailhook:
We need some message like `Compressing the image... This may take a few minutes.` because it looks like hanging.

tailhook:
Also I've waited 4 minutes for image compression, but then got error when uploading. This looks ugly. But I'm not sure we should do anything with this, just noting the issue.

src/launcher/push.rs
```diff
@@ -0,0 +1,87 @@
...
+                .env("container_name", &opt.name)
+                .env("hash", &hash);
```
tailhook: This isn't full hash. We should call it `short_hash` (which refers to `vagga _version_hash --short`), as we may want to make a full hash available later.

anti-social:
Maybe we should change default compression for `_push_image` command?
```
[alexk@localhost:~/projects/vagga/tests/ubuntu]$ time ../../vagga _pack_image -z trusty > trusty.tar.gz
(1/1) Installing alpine-keys (1.1-r0)
OK: 0 MiB in 1 packages
fetch http://distrib-coffee.ipsl.jussieu.fr/pub/linux/alpine/alpine/v3.3/main/x86_64/APKINDEX.tar.gz
(1/2) Installing musl (1.1.12-r3)
(2/2) Installing gzip (1.6-r1)
OK: 0 MiB in 3 packages
Compressing the image... This may take a few minutes.
../../vagga _pack_image -z trusty > trusty.tar.gz  7,00s user 0,33s system 76% cpu 9,585 total
[alexk@localhost:~/projects/vagga/tests/ubuntu]$ time ../../vagga _pack_image -J trusty > trusty.tar.xz
(1/1) Installing alpine-keys (1.1-r0)
OK: 0 MiB in 1 packages
fetch http://mirrors.gigenet.com/alpinelinux/v3.3/main/x86_64/APKINDEX.tar.gz
(1/3) Installing musl (1.1.12-r3)
(2/3) Installing xz-libs (5.2.2-r1)
(3/3) Installing xz (5.2.2-r1)
OK: 0 MiB in 4 packages
Compressing the image... This may take a few minutes.
../../vagga _pack_image -J trusty > trusty.tar.xz  57,01s user 0,81s system 90% cpu 1:04,03 total
[alexk@localhost:~/projects/vagga/tests/ubuntu]$ ls -lh trusty.*
-rw-r--r-- 1 alexk users 168M Мар 23 20:49 trusty.tar
-rw-r--r-- 1 alexk users  58M Мар 23 20:51 trusty.tar.gz
-rw-r--r-- 1 alexk users  39M Мар 23 20:52 trusty.tar.xz
```

lilianmoraru:
Does this feature pack everything you've done with the environment?
What exactly gets packed?
If I did a `BindRW` in vagga, how would that translate when I will give the image to somebody else? They would have to do something akin' to `BindRW` with whatever tool they use(may be vagga, may be not)?
Can these images be loaded by vagga, if yes, how?

If it is what I think it is(from the description and from the discussion I did not understand very well what this is capable of), than it has great potential(great for CI?). This feature probably should have pretty good documentation.

anti-social:
`vagga` already has `_pack_image` command.

This PR introduces `_push_image` command to pack and upload image to the central server. Also you can specify `image-cache-url` option for container and `vagga` will try to download image from that url. Primary goal of this feature is to simplify `vagga` usage for big teams.

tailhook:
@lilianmoraru this command packs exactly what is stored in `~/.vagga/container_name`, without any volumes. So you need original vagga.yaml and original project to use it.

No, it doesn't do anything with `BindRW`. Vagga will mount whatever is in the destination path.

As described above primary use case of the functionality is to cache images for vagga itself (just set an url in the vagga.yaml). You can use it for other things too if you wish, in that case it's barely a shortcut to `vagga _pack_image` + whatever is written in `push-image-script`.

tailhook:
Squash-merged in daf10e3

PR #200, opened 2016-03-17
Ruby support

livioribeiro:
One thing I am not sure about is if we should update rubygems itself during build (currently this is done by default with possibility to disable).

Ubuntu inherited from Debian and forbids updating rubygems unless we tell it to really do it through environment variable.


tailhook:
Looks great. I'll take more careful look soon. Just a quick question: is there any good reason to parse Gemfile for exact packages and versions?

I know we do that for other package managers, because it's easy and non-controversial. So adding a comment or reordering packages doesn't influence package version. But since Gemfile is the code, and it may contain dependencies that are not caught by regexes, it looks like it's better to just hash the whole file. So trade some spurious rebuilds for the correctness in all cases. 

So are there any other Ruby tools, that parse Gemfile without actually evaluating it?

(the heuristics in `requires_git` are fine, I'm just talking about versioning)

livioribeiro:
If some spurious rebuilds are not a problem, then there's no good reason to try to parse the Gemfile.

I did a quick search and found 2 tools that can parse Gemfiles:
 * https://github.com/gemnasium/gemnasium-parser, which actually uses regular expressions;
 * and https://gitlab.com/balasankarc/gemfileparser, ironically written in python;

I believe it is better to just hash the whole file.

livioribeiro:
Gemfile versioning now just hashes the whole file.

I also did the same to the lock file since it might not reliable, for example, setting a gem to be retrieved from a git repo would significantly alter the lock file structure and the algorithm had to take that new structure into account.

tailhook:
Just squashed and merged in 05afab6. Thank you for such valuable contributions!

PR #202, opened 2016-03-18
Add `tags` field to Command and allow filtering processes by tags

popravich:
This allow marking some commands in supervise block with common tag and then running only those
by filtering on that tag:
```yaml
commands:
  run: !Supervise
    children:
      db: !Command
        tags: [db, service]
      redis-master: !Command
        tags: [db, service]
      redis-slave: !Command
        tags: [db, service]
      nginx: !Command
        tags: [service]
      frontend_app: !Command
        tags: [app]
      backend_app: !Command
        tags: [app]
      tasks_queue: !Command
        tags: [app]
```
and then runn all "services" with:
`vagga run --only-tags service`
instead of:
`vagga run --only db redis-master redis-slave nginx` or
`vagga run --exclude frontend_app backend_app tasks_queue`

**no tests yet**

### DIFF
--- a/src/launcher/supervisor.rs
+++ b/src/launcher/supervisor.rs
@@ -39,12 +39,14 @@ pub fn run_supervise_command(settings: &Settings, workdir: &Path,
         let mut ap = ArgumentParser::new();
         ap.set_description(sup.description.as_ref().map(|x| &x[..])
             .unwrap_or("Run multiple processes simultaneously"));
-        ap.refer(&mut only).metavar("PROCESS_NAME")
+        ap.refer(&mut only).metavar("PROCESS_NAME_OR_TAG")
             .add_option(&["--only"], List, "
-                Only run specified processes");
-        ap.refer(&mut exclude).metavar("PROCESS_NAME")
+                Only run specified processes.
+                This matches both names and tags");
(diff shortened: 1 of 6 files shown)


src/launcher/supervisor.rs
```diff
@@ -45,6 +46,9 @@ pub fn run_supervise_command(settings: &Settings, workdir: &Path,
         ap.refer(&mut exclude).metavar("PROCESS_NAME")
             .add_option(&["--exclude"], List, "
                 Don't run specified processes");
+        ap.refer(&mut tags_only).metavar("COMMAND_TAG")
```
tailhook: Could be just "TAG", no need for "COMMAND_"

tailhook:
Do we need an `--only-tags` argument? I mean `--only` can work with tags as well as processes. No need to be more strict that needed. `--exclude` should work with tags too.

popravich:
Tags may match command names and it would be unclear what to run, as in example above this would be ambiguous:
`vagga run --only db`

popravich:
PR updated:
* [x] `--only` & `--exclude` now filter by tags as well;
* [x] tests added;
* [x] docs updated;
* [x] _compgen updated;

PR #207, opened 2016-03-23
PHP/Composer fixes

livioribeiro:
This fixes two problems in php/composer integration:

The first one is that the symlink to the composer executable `/usr/local/bin/composer -> /tmp/composer.phar` wasn't working, calling `composer` always returned "command not found". Now, composer is installed directly to `/usr/local/bin` and removed after container is built.

The second one that the module `php-zip` was not installed by default in Alpine. This is not really a fix, but a convenience for some popular tools, like the laravel installer, that requires this module.


tailhook:
Merged. Thanks!

PR #211, opened 2016-04-01
Small documentation improvements

popravich:
* better `code-block`s
* drop duplicate usage of `what_is_vagga` from examples
* update django/laravel examples with shorter command lines

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -8,7 +8,6 @@ Tutorials
 .. toctree::
    :maxdepth: 1
 
-   what_is_vagga
    examples/tutorials/django
    examples/tutorials/laravel
    examples/tutorials/rails
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -8,6 +8,7 @@ Tutorials
 .. toctree::
    :maxdepth: 1
 
+   what_is_vagga
    examples/tutorials/django
    examples/tutorials/laravel
    examples/tutorials/rails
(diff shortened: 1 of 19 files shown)


docs/examples.rst
```diff
@@ -8,7 +8,6 @@ Tutorials
 .. toctree::
    :maxdepth: 1
 
-   what_is_vagga
```
tailhook: I believe this is fine. Since it is actually a tutorial. And it's the simplest one.

popravich: well, its being shown twice in left panel and its a bit confusing.

popravich: Maybe a reference to that doc would work

tailhook: Well, I don't think it's a big deal. If you find a better way to place link there, do it. Currently, it's more helpful than distracting.

popravich: ok, how about this -- 1f65ceec6f721a31fbce9f23e73e586153f591b3?

tailhook:
> ok, how about this -- 1f65cee?

So rather than duplicating a link, you duplicated whole content? I mean it's probably fine either way. I just don't see an exact problem you're fixing. Link in two places is okay. we have cross-links all over the documentation. Is it wrong submenu opened in left bar when you navigate to a page?

While, search engines don't like duplicated content, it's not my main concern. My main concern is that `/info.html` now contains both: the links to the subpages and long tutorial on itself. I.e. it serves as both: the index page and text page. It is usually confusing. If I see a list of links at the top of the page with lots of text below I think that links above refer to the parts of this page.

tailhook:
Merged everything except `what_is_vagga` changes for now. Thanks!

PR #215, opened 2016-04-04
Remove gettimeofday

### LINKED ISSUE #153: Get rid of gettimeofday extern function
This is from pre-cargo vagga. Should get rid of `get_time` and `Time` type too. Replace with "time" crate instead

livioribeiro:
Closes #153


tailhook:
Looks fine. Thanks!

PR #216, opened 2016-04-04
Add Unzip command

anti-social:


src/builder/unzip.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub fn unzip_file(_ctx: &mut Context, src: &Path, dst: &Path, subdir: &Path)
+    -> Result<(), String>
+{
+    let file = try_msg!(File::open(src),
+        "Cannot open archive: {err}");
+    let mut zip = try_msg!(ZipArchive::new(file),
+        "Error when unpacking zip archive: {err}");
+
+    // TODO check sha256 sum
```
tailhook: Well I you add an option, please implement it. It's absent for Tar command is just because it was far far ago, when we didn't have cargo and so a good library that implements this thing. Now it's easy to implement.

tailhook:
Other than sha256 I don't see any issues with this.

tailhook:
And probably `unzip` should be in `builder/commands` I think it's just an old mistake to keep `tarcmd` out of the `commands`

anti-social:
Ok I'll do this.

 I wanted to implement checking sha256 in another PR for all commands that have this option.

tailhook:
Sorry, that I'm skipped details in previous comment. Here is how hashsum changes the game:

1. Since we have cryptographically strong hash sum of the download anyway we don't need to version URL of the download
2. It's better to make cache content-addressable. I.e. make cache entry derived from sha256sum rather than URL (we should probably add another cache folder)

This is to make useful improvement: you can change a URL to a mirror, proxy, local dir, whatever, and this doesn't change the container. At some point in future we can also make something similar to `url.<base>.insteadOf` setting of git.

I think it's important to get checksumming right form the start because changing checksums of the containers is painful for users.

tailhook:
Well, and another issue: you must download a file to a temporary location, check the checksum, and then move the file into a permanent place (and delete otherwise). We don't want to store broken downloads in the cache (when reading old file from the cache it should be checked too, and double-checking of the freshly fetched files should be avoided, for that matter)

tailhook:
Merged. Thanks!

PR #222, opened 2016-04-13
PHP/Composer improvements

livioribeiro:
This PR adds the following improvements:

* Composer binary is cached and updated using its self-update feature
* Add option to keep composer binary after build
* Add option to change the composer vendor directory


tailhook:
Merged. Thanks!

PR #226, opened 2016-04-24
Permissions for unzip command

anti-social:
Could you please add executable file into `test-file.zip`:

```
$ cat install.sh
#/bin/sh
PREFIX=${PREFIX:-/opt}
install -d $PREFIX
cp -r dir $PREFIX
```

### DIFF
--- a/tests/generic.bats
+++ b/tests/generic.bats
@@ -190,25 +190,30 @@ setup() {
     run vagga _build unzip-local
     printf "%s\n" "${lines[@]}"
     link=$(readlink .vagga/unzip-local)
-    [[ $link = ".roots/unzip-local.24255fd9/root" ]]
+    [[ $link = ".roots/unzip-local.6e26085b/root" ]]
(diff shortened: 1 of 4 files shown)


tests/generic.bats
```diff
@@ -193,16 +193,21 @@ setup() {
     [[ $link = ".roots/unzip-local.24255fd9/root" ]]
     [[ $(cat .vagga/unzip-local/root/test/1/dir/file.txt) = "Hello" ]]
     [[ $(cat .vagga/unzip-local/root/test/1/dir/file2.txt) = "2" ]]
+    # [[ -x .vagga/unzip-local/root/test/1/install.sh ]]
```
tailhook: Why is it commented out?

anti-social: Because I need `install.sh` in test zip archive otherwise this test will fail.

tailhook:
Uploaded. But for the future:

1. Please mention somewhere in the pull request that it is *not ready for merge*
2. I'm not sure why did you put commented out test. You will fix it anyway. It should fail for now and that's fine.

PR #229, opened 2016-04-25
Check minimum vagga version before validating config

anti-social:


tailhook:
Looks fine, except you shouldn't parse configuration again. Use already parsed AST.

anti-social:
Probably it is worth to add a function into `rust-quire` that accepts AST and validate it? Otherwise there will be too much copy-paste.

tailhook:
> Probably it is worth to add a function into rust-quire that accepts AST and validate it?

I'm not sure this is exact semantic it should have. But yes, the whole idea is to split `parse_config` into smaller parts. That was always intended to be a quick shortcut, until you have more specific needs.

Just got another thought: it may be worth to implement version check as a validator. Since the validators are executed at the AST stage, it may work as good as this hack.

anti-social:
Can't understand why `gem bundler lock` test fails. It passes on my local machine.

tailhook:
I don't know either. Probably error is somewhat transient.

Merged. Thanks!

PR #238, opened 2016-05-07
Add --no-remote to firefox doc example, switch to xenial and other minor doc fixes

lilianmoraru:
The identation for `!Install` might look weird in `vagga_webgl_intel.yaml` but that's actually how it has to be aligned in order for the generated docs identation to look nice(not a whitespace issue).


tailhook:
Merged. Thanks!

Issue #241, opened 2016-05-10
Implement `!RunAs` builder


tailhook:
The proposed syntax is:
```
- !RunAs
  user-id: 1
  external-user-id: 0
  group-id: 0
  supplementary-groups: [1,2,3]
  script: |
    echo hello
```
With only `script` being mandatory field.

This should make issues like #71 avoidable even on old systems.

tailhook:
Done in #251 / ae38ddb

PR #251, opened 2016-05-17
RunAs command (merge after uidmap)

popravich:
Implements #241 

Done:
* [x] code
* [x] tests
* [x] docs
* [ ] wait `uidmap` branch merged into master

### DIFF
--- a/src/builder/commands/generic.rs
+++ b/src/builder/commands/generic.rs
@@ -6,7 +6,7 @@ use quire::validate as V;
 
 use super::super::context::Context;
 use super::super::super::path_util::ToRelative;
-use process_util::{capture_stdout};
+use process_util::{capture_stdout, set_fake_uidmap};
 use build_step::{BuildStep, VersionError, StepError, Digest, Config, Guard};
 
 // Build Steps
@@ -27,6 +27,26 @@ tuple_struct_decode!(Cmd);
 impl Cmd {
     pub fn config() -> V::Sequence<'static> {
         V::Sequence::new(V::Scalar::new())
+    }
+}
+
+#[derive(RustcDecodable, Debug)]
+pub struct RunAs {
+    pub user_id: u32,
+    pub group_id: u32,
+    pub supplementary_gids: Vec<u32>,
+    pub external_user_id: Option<u32>,
+    pub script: String,
+}
+
+impl RunAs {
+    pub fn config() -> V::Structure<'static> {
+        V::Structure::new()
+        .member("user_id", V::Numeric::new().min(0).default(0))
+        .member("group_id", V::Numeric::new().min(0).default(0))
+        .member("external_user_id", V::Numeric::new().min(0).optional())
+        .member("supplementary_gids", V::Sequence::new(V::Numeric::new()))
+        .member("script", V::Scalar::new())
     }
 }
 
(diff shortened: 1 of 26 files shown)


src/builder/commands/generic.rs
```diff
@@ -262,3 +290,47 @@ impl BuildStep for Env {
...
+    {
+        hash.text("user_id", &self.user_id);
+        hash.text("group_id", &self.group_id);
+        if let Some(euid) = self.external_user_id {
+            hash.text("external_user_id", &euid);
+        }
+        for i in self.supplementary_gids.iter() {
+            hash.text("supplementary_gids", &i);
+        }
+        hash.field("RunAs", &self.script);
```
tailhook: s/RunAs/script/

src/builder/commands/generic.rs
```diff
@@ -77,9 +97,9 @@ fn find_cmd<P:AsRef<Path>>(ctx: &Context, cmd: P)
         "-- empty PATH --".to_string()));
 }
 
-pub fn run_command_at_env(ctx: &mut Context, cmdline: &[String],
+fn build_command(ctx: &mut Context, cmdline: &[String],
```
tailhook: All this old stuff is structured like this from pre-`unshare` times, when we had no `Command` structure.

I mean you can make a function that makes some common settings and returns command. But you shouldn't add the arguments that are already there on current interface: `cmdline`, `env`, `path` (latter is essentially a workdir). This would make code nicer and even avoid some memory allocations.

tailhook:
Other than that, looks fine. Shouldn't we also add a workdir?

popravich:
Updated, please review

src/builder/commands/generic.rs
```diff
@@ -303,19 +300,18 @@ impl BuildStep for RunAs {
...
         -> Result<(), StepError>
     {
         if build {
-            let mut cmd = try!(build_command(&mut guard.ctx,
-                                             &["/bin/sh".to_string(),
-                                               "-exc".to_string(),
-                                               self.script.to_string()],
-                                             &Path::new("/work"),
-                                             &[]));
+            let mut cmd = Command::new(&"/bin/sh".to_string());
```
tailhook: No need for `to_string()` as far as I remember. That's how magic works in `File::open` and similar things.

src/builder/commands/generic.rs
```diff
@@ -303,19 +300,18 @@ impl BuildStep for RunAs {
...
+            setup_command(&guard.ctx, &mut cmd);
+            cmd.args(&["-exc".to_string(), self.script.to_string()]);
```
tailhook: This is similar, you don't need `to_string()` conversion, which is allocating a `String`. and you don't need to create a slice, just:
```
cmd.arg("-exc");
cmd.arg(&self.script)
```

(The slice is cost-less, but using separate call allows to use different type for each argument, if needed)

src/builder/commands/generic.rs
```diff
@@ -303,19 +300,18 @@ impl BuildStep for RunAs {
...
+            cmd.current_dir(&Path::new("/work"));
```
tailhook: This is the same, you don't need `Path::new` any more.

tailhook:
Squash-merged in in ae38ddb

popravich:
I think we should add `workdir` parameter. Will make a new PR

PR #252, opened 2016-05-18
add workdir option to RunAs build step

popravich:


tailhook:
1. We have `work-dir` (with a dash) in command, so should replicate it here
2. We allow relative dir (say `docs`) in command, so I don't see any reason why we would prohibit it here

popravich:
done

src/builder/commands/generic.rs
```diff
@@ -311,7 +315,11 @@ impl BuildStep for RunAs {
             setup_command(&guard.ctx, &mut cmd);
             cmd.arg("-exc");
             cmd.arg(&self.script);
-            cmd.current_dir("/work");
+            if self.work_dir.is_absolute() {
```
tailhook: You don't need this check, it's handled by `.join()` (https://is.gd/lIK4Fx)

popravich:
fixed

tailhook:
Merged. Thanks!

Issue #269, opened 2016-06-05
Running vagga in docker for mac


tailhook:
Assuming you have a docker set up, you can run vagga in docker with the
following command-line:

    docker run --volume=$(pwd):/work --workdir=/work --privileged \
        -i -t --rm tailhook/vagga:latest \
        /vagga/bin/vagga --ignore-owner-check whateever vagga command

To make it easier to use, you should add the alias to your `~/.bashrc`:

    alias='docker run --volume=$(pwd):/work --workdir=/work --privileged \
        -i -t --rm tailhook/vagga:latest \
        /vagga/bin/vagga --ignore-owner-check whateever vagga'

Then you can run vagga just by running as always:

    vagga command

However, this will probably not work in docker for the following reasons (simiarly to vagrant issues):

1. [ ] Filesystem in .vagga might be not support hardlinks, or it might not be efficient to store images directly in host system (which we do when mount `/work`)
2. [ ] Images will not be cached, we need to figure out a good place to put container caches (can we make a volume for that?) should it be the data container or something like this?
3. [ ] We need to figure out userid which owns files inside the docker for mac container, so we can remove `--ignore-owner-check` (on the other side I'm assuming that even files owned by root will be written as normal user in OS X)
4. [ ] Any other usability issues (hanging processes? leftover images?)
5. [ ] Any other issues

lilianmoraru:
`alias vagga=...`

tailhook:
Sure. Thanks!

tailhook:
Notes to myself:

1. Alpine mostly works. But we have issues with setuid processes:
```
$ docker run --volume=$(pwd):/work --workdir=/work --privileged -i -t --rm tailhook/vagga:latest /vagga/bin/vagga --ignore-owner-check server
(1/1) Installing alpine-keys (1.1-r0)
OK: 0 MiB in 1 packages
fetch http://mirror.leaseweb.com/alpine/v3.2/main/x86_64/APKINDEX.tar.gz
(1/15) Installing musl (1.1.11-r2)
(2/15) Installing busybox (1.23.2-r0)
ERROR: Failed to set ownership on bin/bbsuid.apk-new: Permission denied
ERROR: Failed to preserve modification time on bin/bbsuid.apk-new: Permission denied
Executing busybox-1.23.2-r0.post-install
(3/15) Installing alpine-baselayout (2.3.2-r0)
Executing alpine-baselayout-2.3.2-r0.pre-install
[...]

Issue #53, opened 2015-08-29
Running vagga on strange file systems.


tailhook:
The examples are:

1. Running on vagrant in `/vagrant` folder
2. Same as (1) but running on windows
3. Fuse file systems
4. Some kind of NFS
5. Fat32 or some other weird file system (not sure it's important)

In the case (2) and (5) are problematic because they doesn't support symlinks. Case (1) and (3) may have issue with user namespaces (there is a patch in vagrant which probably solve the issue), because they mount filesystem with `user=x,group=x`. Not sure about NFS. Not sure about OS X.

There are three kind of issues:

1. Unable to write files inside the namespace (at least for case (1) above)
2. Unable to do symlinks
3. Inefficient resource usage (NFS, Fuse, maybe vboxsf)

The case (3) is currently fixed with [storage-dir in settings](http://vagga.readthedocs.org/en/latest/settings.html)

For avoiding other issues the current best practice is:
```
mkdir .vagga
sudo mount --bind /some/writable/folder .vagga
vagga whatever
```

It's probably possible to:

1. Automatically detect this condition to vagga and propose `storage-dir: /home/<user>/.cache/vagga`
2. Setup the storage dir in a way that doesn't require manipulating .vagga folder inside the user namespace and doesn't require symlinks in source folder

The problems to solve:

1. There are links `.vagga/<container_name>` which are super-useful in the IDE's, it's not clear how to setup them in all the cases above (i.e. when symlinks doesn't work)
2. The layout of `.vagga` folder to support this. Probably `.vagga/.lnk` might be a text file
3. Security measures (i.e. if we manipulate `.vagga` before setting up chroot, we may touch folder outside of both project dir and `storage-dir`, which is a security issue)

PR #271, opened 2016-06-10
Document errors with shared folders

skade:
This is an issue I found while trying out vagga trough OS X.

It's rather obvious in hindsight, but hey :).

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -109,3 +109,25 @@ Both cases should inhibit the warning automatically, but as a last resort
 you may try ``vagga --ignore-owner-check``. If you have good case where this
 works, please file an issue and we might make the check better.
 
+.. _virtualbox:
+
+Don't run vagga on shared folders
+---------------------------------
+
+If you are testing or using vagga using an OS X machine, you might be tempted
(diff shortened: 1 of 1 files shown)


tailhook:
Hi, looks nice addition. It's known thing but have never been placed in documentation. I have a few questions though:

1. Does setting `storage-dir` actually works on mac? As far as I remember it doesn't work on windows, because vagrant refuses to make symlinks in vagrant directory (i.e. we do them outside of `storage-dir`), so current best practice is something like this (documented in #53):
```
mkdir .vagga
sudo mount --bind /some/writable/folder .vagga
```

2. Usually, people are happy with `vagrant vagga` plugin which does that automatically. So do you think it's better to recommend a plugin, a `mount --bind` hack or `storage-dir`?

----

If you have a little more time it would be helpful to test vagga in docker for mac (#269) it would probably need same hacks, like mounting `.vagga` into linux file system and so on. But it has two important improvements:

1. It declares the support of inotify (so you can use auto-rebuild, hot-reload and similar techniques)
2. You don't have to run a separate virtual machine for each project you have a vagga.yaml for

----

Thanks, for your help. We definitely want to make vagga easier for OS X users.

skade:
Hi Paul,

I didn't even know about vagrant-vagga :). I will revise the section over the weekend.

Best,
Florian

tailhook:
Anyway, let me merge it as is and we can add more info later.

PR #272, opened 2016-06-10
Remove need to build musl and llvm to build vagga

skade:
This removes the need to build musl and llvm, as
no custom rust build is needed anymore. Musl GCC
is available from package repositories.

Also splits the build of the base container from
the build of the rust container, giving a good
breakpoint while working on changes.

Last part made it much easier for me to develop on
a flaky connection on a train ;).

Also fixes the problem that static.rust-lang.org is not
available though HTTP anymore.

This is definitely a bit rough, I will thoroughly test it later.

Also, on a sidenote, I ignored your readme and vagga now
builds just fine with rustup and the appropriate target.

### DIFF
--- a/vagga.yaml
+++ b/vagga.yaml
@@ -16,33 +16,16 @@ containers:
     - !UbuntuUniverse
     - !Install [build-essential, ca-certificates]
     - !Install [file]  # dependency of checkinstall (bug #46)
-    - !BuildDeps [cmake, libc++-dev, curl, autoconf, automake, libtool]
-    # Build script derived from
-    # https://github.com/rust-lang/rust-buildbot/blob/master/slaves/linux/build-musl.sh
-    - !TarInstall
-      url: http://www.musl-libc.org/releases/musl-1.1.14.tar.gz
-      script: |
-        ./configure --prefix=/musl --disable-shared
-        make -j2
-        make install
 
-    - !Tar
-      url: http://llvm.org/releases/3.8.0/llvm-3.8.0.src.tar.xz
-      path: /tmp/llvm
     - !TarInstall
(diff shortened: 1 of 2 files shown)


vagga.yaml
```diff
@@ -6,43 +6,30 @@ containers:
     - !Install [alpine-base, py-sphinx, make]
     - !Py2Requirements docs/requirements.txt
 
+  ubuntu-base:
+    setup:
+    - !Ubuntu xenial
+    - !UbuntuUniverse
+    - !Install [build-essential, ca-certificates]
+    - !Install [file]  # dependency of checkinstall (bug #46)
+    - !Install [openssl] # necessary as downloads from static.rust-lang.org need SSL now
```
tailhook: The Tar and TarInstall commands don't use the target system for downloading. It's done by setting intermediary alpine system which has wget and ca-certificates. So you don't need this I think.

tailhook:
> Last part made it much easier for me to develop on
a flaky connection on a train ;).

I think it's fine to do this temporarily, but not to commit. Subcontainer usage has some cost and there is no evidence that base container will be updated less frequently than the full one (i.e. updating it once every 6 weeks is not a problem).

> Also, on a sidenote, I ignored your readme and vagga now builds just fine with rustup and the appropriate target.

Which part of readme? We should probably update then.

----

Overall, I like the changes. But I'm going to release 0.6 in a day or two, and I think I should postpone the pull request to 0.6.1.

skade:
See the updated file: yes, the need for openssl vanishes when you use one container. Horray, sideffects :). Before that, wget complained about not being able to load `ssl_helper`.

I also updated the piece of docs I was referring to.

tailhook:
Merged. Thanks!

`vagga cached-make` is not needed anymore too, since we stopped to build rust itself. But I'll fix it shortly.

PR #274, opened 2016-06-12
Use SHA2 crate instead of Shaman

### LINKED ISSUE #254: Use sha2 crate instead of shaman

leonardoce:
This patch removes the usage of the Shaman crate, using instead at his place.

Closes #254


tailhook:
Merged. Thanks!

PR #275, opened 2016-06-13
Remove time crate in favor of std::time

skade:
This ports all uses of the time crate over
to equivalents in std::time.

Note that this currently relies on a fork of the signal crate.

### DIFF
--- a/src/builder/timer.rs
+++ b/src/builder/timer.rs
@@ -3,7 +3,7 @@ use std::io::Write;
 use std::path::Path;
 use std::fs::File;
 use std::fmt::{Arguments};
-use std::time::{Instant, Duration};
+use std::time::{Instant, Duration, SystemTime, UNIX_EPOCH};
 
 const NANO_FACTOR: f64 = 0.000000001;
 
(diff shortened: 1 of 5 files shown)


src/builder/timer.rs
```diff
@@ -3,42 +3,39 @@ use std::io::Write;
...
+        let now = Instant::now();
         let mut res = TimeLog {
             file: try!(File::create(path)),
-            start: tm,
-            prev: tm,
+            start: now,
+            prev: now,
         };
-        try!(res.mark(format_args!("Start {}", time_as_f64(tm))));
+        try!(res.mark(format_args!("Start {:?}", now)));
```
tailhook: Not sure about this one. It used to be a unix timestamp. Given that it's easy to convert instant to a unixtime, i think we should keep it that way.

tailhook:
Just uploaded signal 0.2 to crates.io

tailhook:
Any chance you get this one finished? Or should I steal the PR?

skade:
Hi Paul,

thanks for the reminder, I will probably get that finished over the weekend.

We'll need to switch away from Instant, btw, as it cannot be converted to a timestamp easily :).

Best,
Florian

tailhook:
Hm, yes. But you just need SystemTime to show start timestamp, everything else may work with Instant.

skade:
Finally got around to fixing this. :)

tailhook:
Merged. Thanks!

PR #278, opened 2016-06-13
Hashing of included pip requirements

lud4ik:
What does it mean?
```
src/builder/commands/pip.rs:331:25: 332:74 warning: unused result which must be used, #[warn(unused_must_use)] on by default
src/builder/commands/pip.rs:331                         version_req(hash,
src/builder/commands/pip.rs:332                             fname.parent().unwrap().join(req).as_path());

```

### DIFF
--- a/src/builder/commands/pip.rs
+++ b/src/builder/commands/pip.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::io::{BufReader, BufRead};
 use std::path::{Path, PathBuf};
 
@@ -329,8 +329,21 @@ fn parse_req_filename(line: &str) -> Option<&str> {
     return None;
 }
 
-fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
-    let path = Path::new("/work").join(fname);
+fn version_req(hash: &mut Digest, fname: &Path, used: &mut HashSet<String>) ->
+               Result<(), VersionError> {
+
+    let path = try!(Path::new("/work")
+                       .join(fname)
+                       .canonicalize()
+                       .map_err(|e| VersionError::Io(e, fname.to_path_buf())));
+
+    let name = format!("{:?}", path);
+    if used.contains(&name[..]) {
+        return Ok(())
+    }
+
+    used.insert(name);
+
     let f = try!(File::open(&path)
                       .map_err(|e| VersionError::Io(e, path.clone())));
 
@@ -343,7 +356,9 @@ fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
             continue;
         }
         if let Some(req) = parse_req_filename(chunk) {
-            try!(version_req(hash, &fname.parent().unwrap().join(req)));
+            try!(version_req(hash,
+                             &fname.parent().unwrap().join(req),
+                             used));
         }
         // Should we also ignore the order?
         hash.item(chunk);
@@ -355,7 +370,7 @@ impl BuildStep for Py2Requirements {
     fn hash(&self, _cfg: &Config, hash: &mut Digest)
         -> Result<(), VersionError>
     {
-        version_req(hash, &self.0)
+        version_req(hash, &self.0, &mut HashSet::new())
     }
     fn build(&self, guard: &mut Guard, build: bool)
         -> Result<(), StepError>
@@ -376,7 +391,7 @@ impl BuildStep for Py3Requirements {
     fn hash(&self, _cfg: &Config, hash: &mut Digest)
         -> Result<(), VersionError>
     {
-        version_req(hash, &self.0)
+        version_req(hash, &self.0, &mut HashSet::new())
     }
     fn build(&self, guard: &mut Guard, build: bool)
         -> Result<(), StepError>
--- a/src/builder/commands/pip.rs
+++ b/src/builder/commands/pip.rs
@@ -335,7 +335,8 @@ fn version_req(hash: &mut Digest, fname: &Path, used: &mut HashSet<String>) ->
 
     let name = format!("{:?}", path);
     if used.contains(&name[..]) {
-        return Ok(())
+        return Err(VersionError::String(
+            format!("Cyclic requirement: {}", name)))
     }
 
     used.insert(name);
@@ -355,6 +356,7 @@ fn version_req(hash: &mut Digest, fname: &Path, used: &mut HashSet<String>) ->
             try!(version_req(hash,
                              &fname.parent().unwrap().join(req),
                              used));
+            continue;
         }
         // Should we also ignore the order?
         hash.item(chunk);
--- a/src/builder/commands/pip.rs
+++ b/src/builder/commands/pip.rs
@@ -1,4 +1,5 @@
 use std::fs::File;
+use std::collections::HashMap;
 use std::io::{BufReader, BufRead};
 use std::path::{Path, PathBuf};
 
@@ -314,14 +315,18 @@ impl BuildStep for Py3Install {
 }
 
 fn parse_req_filename(line: &str) -> Option<&str> {
-    if line.starts_with("-r") {
-        return Some(line[2..].trim())
-    } else if line.starts_with("--requirement ") ||
-              line.starts_with("--requirement=") {
-        return Some(line[14..].trim())
-    } else {
-        return None
+    let res =
+        vec!["-r", "--requirement ", "--requirement=",
+             "-c", "--constraint ", "--constraint="]
+            .into_iter()
+            .map(|x| (x, x.len()))
+            .collect::<HashMap<_, _>>();
+    for (prefix, pos) in res {
+        if line.starts_with(prefix) {
+            return Some(line[pos..].trim());
+        }
     }
+    return None;
 }
 
 fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
--- a/src/builder/commands/pip.rs
+++ b/src/builder/commands/pip.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::io::{BufReader, BufRead};
 use std::path::{Path, PathBuf};
 
@@ -315,15 +315,11 @@ impl BuildStep for Py3Install {
 }
 
 fn parse_req_filename(line: &str) -> Option<&str> {
-    let res =
-        vec!["-r", "--requirement ", "--requirement=",
-             "-c", "--constraint ", "--constraint="]
-            .into_iter()
-            .map(|x| (x, x.len()))
-            .collect::<HashMap<_, _>>();
-    for (prefix, pos) in res {
+    let res = vec!["-r", "--requirement ", "--requirement=",
+                   "-c", "--constraint ", "--constraint="];
+    for prefix in res.iter() {
         if line.starts_with(prefix) {
-            return Some(line[pos..].trim());
+            return Some(line[prefix.len()..].trim());
         }
     }
     return None;
--- a/src/builder/commands/pip.rs
+++ b/src/builder/commands/pip.rs
@@ -313,24 +313,37 @@ impl BuildStep for Py3Install {
     }
 }
 
+fn parse_req_filename(line: &str) -> Option<&str> {
+    if line.starts_with("-r") {
+        return Some(line[2..].trim())
+    } else if line.starts_with("--requirement ") ||
+              line.starts_with("--requirement=") {
+        return Some(line[14..].trim())
+    } else {
+        return None
+    }
+}
+
 fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
     let path = Path::new("/work").join(fname);
-    let err = |e| VersionError::Io(e, path.clone());
-    File::open(&path)
-    .and_then(|f| {
-            let f = BufReader::new(f);
-            for line in f.lines() {
-                let line = try!(line);
-                let chunk = line[..].trim();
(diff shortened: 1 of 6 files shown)


tailhook:
It means that `version_req` returns a `Result<_, SomeError>` which means you must process error somehow, for example, propagate it using `try!`

src/builder/commands/pip.rs
```diff
@@ -326,6 +326,12 @@ fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
                 if chunk.len() == 0 || chunk.starts_with("#") {
                     continue;
                 }
+                if chunk.starts_with("-r ") {
```
tailhook: I don't think this parsing is good enough. You need to parse 4 cases I think:
`-rXXX`, `-r XXX`, `--requirement XXX`, `--requirement=XXX`.

src/builder/commands/pip.rs
```diff
@@ -326,6 +326,12 @@ fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
...
+                    if let Some(req) = chunk.rsplit(" ").next() {
+                        version_req(hash,
+                            fname.parent().unwrap().join(req).as_path());
```
tailhook: It's usually enough `&something.join(x)` instead of `somethign.join(x).as_path()`

src/builder/commands/pip.rs
```diff
@@ -326,11 +341,18 @@ fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
...
+                if let Some(req) = parse_req_filename(chunk) {
+                    included_files.push(req);
+                }
                 // Should we also ignore the order?
                 hash.item(chunk);
             }
             Ok(())
-    }).map_err(err)
+    }).map_err(err));
+    for req in &included_files {
```
tailhook: Any specific reason you don't do this recursively, but rather use a list? If there is, it's cute to insert it into the code as a comment.

lud4ik: I couldn't make a recursive call in loop because of:
```
   Compiling vagga v0.6.0 (file:///work)
<std macros>:6:1: 6:32 error: the trait bound `std::io::Error: std::convert::From<version::error::Error>` is not satisfied [E0277]
<std macros>:6 $ crate:: convert:: From:: from ( err ) ) } } )
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/builder/commands/pip.rs:331:25: 332:67 note: in this expansion of try! (defined in <std macros>)
<std macros>:6:1: 6:32 help: run `rustc --explain E0277` to see a detailed explanation
<std macros>:6:1: 6:32 help: the following implementations were found:
<std macros>:6:1: 6:32 help:   <std::io::Error as std::convert::From<xz2::stream::Error>>
<std macros>:6:1: 6:32 help:   <std::io::Error as std::convert::From<nix::Errno>>
<std macros>:6:1: 6:32 help:   <std::io::Error as std::convert::From<nix::Error>>
<std macros>:6:1: 6:32 help:   <std::io::Error as std::convert::From<zip::result::ZipError>>
<std macros>:6:1: 6:32 help: and 4 others
<std macros>:6:1: 6:32 note: required by `std::convert::From::from`
error: aborting due to previous error
``` 
Is there a way to fix it differently? I don't think it should be commented.

tailhook: Sure, this has nothing to do with the recursion, but with the error reporting.

I.e. you need to turn all errors into VersionError instead of io::Error.

So you move `map_err(err)` after `and_then` to before `and_then`. Then you need to wrap `try!(line)` into the `try!(line.map_err(err))`. Then you can `try!(version_req(...))` without the issues because error type is unified everywhere.

lud4ik: For me this variant is more readable (only minus is dublicating of map_err fn):
```
fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
    let path = Path::new("/work").join(fname);
    let f = try!(File::open(&path)
                      .map_err(|e| VersionError::Io(e, path.clone())));

    let f = BufReader::new(f);
    for line in f.lines() {
        let line = try!(line.map_err(|e| VersionError::Io(e, path.clone())));
        let chunk = line[..].trim();
        // Ignore empty lines and comments
        if chunk.len() == 0 || chunk.starts_with("#") {
            continue;
        }
        if let Some(req) = parse_req_filename(chunk) {
            try!(version_req(hash, &fname.parent().unwrap().join(req)));
        }
        // Should we also ignore the order?
        hash.item(chunk);
    }
    Ok(())
}
```

than this style
```
fn version_req(hash: &mut Digest, fname: &Path) -> Result<(), VersionError> {
    let path = Path::new("/work").join(fname);
    let err = |e| VersionError::Io(e, path.clone());
    let mut included_files: Vec<String> = Vec::new();
    try!(File::open(&path)
         .and_then(|f| {
            let f = BufReader::new(f);
            for line in f.lines() {
                let line = try!(line);
                let chunk = line[..].trim();
                // Ignore empty lines and comments
                if chunk.len() == 0 || chunk.starts_with("#") {
                    continue;
                }
                if let Some(req) = parse_req_filename(chunk) {
                    included_files.push(req.to_string());
                }
                // Should we also ignore the order?
                hash.item(chunk);
            }
            Ok(())
    }).map_err(err));
    for req in &included_files {
        try!(version_req(hash, &fname.parent().unwrap().join(req)));
    }
    Ok(())
}
```
I can change it if you like.

lud4ik:
Maybe validate path better? `work/../../../`

src/builder/commands/pip.rs
tailhook: Yes the first one is fine.

tailhook:
> Maybe validate path better? work/../../../req.txt

Don't need to. We do versioning inside the container, so you'll go nowhere when using that path. And you actually can't enforce filesystem limitations by validating path, because there are symlinks.

tailhook:
Now the code looks fine. But please add some minimal unit tests.

lud4ik:
Couldn't check locally
```
+ cd bats
+ ./install.sh /usr
Installed Bats to /usr/bin/bats
git set to automatically installed.
wget set to automatically installed.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  git git-man liberror-perl libexpat1 libssl1.0.0 wget
0 upgraded, 0 newly installed, 6 to remove and 15 not upgraded.
After this operation, 30.4 MB disk space will be freed.
(Reading database ... 10140 files and directories currently installed.)
Removing git (1:2.7.4-0ubuntu1) ...
Removing git-man (1:2.7.4-0ubuntu1) ...
Removing liberror-perl (0.17-1.2) ...
Removing libexpat1:amd64 (2.1.0-7ubuntu0.16.04.1) ...
Removing wget (1.17.1-1ubuntu1) ...
Removing libssl1.0.0:amd64 (1.0.2g-1ubuntu4.1) ...
Processing triggers for libc-bin (2.23-0ubuntu3) ...
ERROR:vagga::version: Error versioning command Step(Depends("vagga")): Error reading "/work/vagga": No such file or directory (os error 2)
ERROR:vagga::wrapper: Error executing _build: Versioner exited exited with code 1
Command <Command "/proc/self/exe" ("vagga_wrapper") "_build" "test"; environ: {"TERM"="xterm","_VAGGA_HOME"="/home/lud4ik","RUST_LOG"="warn",}; uid_map=[UidMap { inside_uid: 0, outside_uid: 1000, count: 1 }, UidMap { inside_uid: 1, outside_uid: 100000, count: 65535 }]; gid_map=[GidMap { inside_gid: 0, outside_gid: 1000, count: 1 }, GidMap { inside_gid: 1, outside_gid: 100000, count: 65535 }]> exited with code 124
```
Is there a way to run only one test from file or just all tests in file?

tailhook:
`vagga test tests/pip.bats`

I'm not sure how you can run a single test, but you may look at [bats documentation](https://github.com/sstephenson/bats)

tests/pip/reqs-included.txt
```diff
@@ -0,0 +1,3 @@
+-r temp1.txt
+--requirement=temp2.txt
+nltk==3.2.1
```
tailhook: Wow! nltk is not the smallest package for the test. I think you should pick the package that already used in some container, just put it in another requirements file and require it from there. You don't even need to install two packages. The upside is that package is already cached in pip cache, so the test will be fast.

lud4ik: I tested only hash diff on change of included file. I'll add build too

tailhook: We don't test hashes because it's impossible to say whether a hash is right. We test some command in the container to find out whether a package is installed. And yes, we compare hash, just to make sure that they don't change when changing the source code of vagga itself.

lud4ik: I didn't tested whether a hash is right. I tested the original issue - to check that after changing included req hash of container changes, too. I used hash values because it was done everywhere, so thought that it was not a big deal. In last commit I changed it to be right: check it without hardcoding hashes.

lud4ik:
It failed in another test. 
```
not ok 43 gem/bundler: GemBundle lock
# (in test file tests/gem_bundler.bats, line 119)
#   `[[ $link = ".roots/bundle-lock.03e2d1cb/root" ]]' failed
```

tailhook:
> It failed in another test.

That one always fails in travis, but doesn't fail either locally or on buildbot. I don't know why.

popravich:
There is also `-c/--constraint` option which may be present in `requirements.txt`, which is basically works as `-r/--requirement`.
Also what about recursive inclusions:
```
# req1.txt
-r req2.txt
some_package

# req2.txt
-r req1.txt
```

lud4ik:
> There is also -c/--constraint option
Give me the doc where it's written.
> what about recursive inclusions
Actually, version_req is recursive.

It seems that I fixed all remarks.

popravich:
https://pip.pypa.io/en/stable/user_guide/#constraints-files

popravich:
>> what about recursive inclusions

> Actually, version_req is recursive.

I mean that you'll end up with stack overflow if your code does not check cyclic (=recursive) dependencies.

tests/pip.bats
```diff
@@ -134,3 +134,21 @@ setup() {
     [[ $status = 0 ]]
     [[ ${lines[${#lines[@]}-1]} = $'pty_copy\r' ]]
 }
+
+@test "py3: recursive requirements files" {
+    echo 'urp' > /work/tests/pip/temp1.txt
```
tailhook: Don't write files in the test dir, unless there is no way out.  You need just two one line files committed to git:
```
include-short.txt: -r requirements.txt
include-nested.txt: --requirement include-short.txt 
```
This assumes files are in `tests/pip` and use `tests/pip/requirements.txt` that contains 'urp'. And you run urp as in all other tests to ensure it's installed.

vagga.yaml
```diff
@@ -121,6 +121,7 @@ commands:
   test: !Command
     description: Run self tests
     container: test
+    prerequisites: [make]
```
tailhook: This one doesn't work. And is explicitly [documented as such](http://vagga.readthedocs.io/en/latest/commandline.html#multiple-commands)

src/builder/commands/pip.rs
```diff
@@ -313,24 +314,41 @@ impl BuildStep for Py3Install {
     }
 }
 
+fn parse_req_filename(line: &str) -> Option<&str> {
+    let res =
+        vec!["-r", "--requirement ", "--requirement=",
+             "-c", "--constraint ", "--constraint="]
+            .into_iter()
+            .map(|x| (x, x.len()))
+            .collect::<HashMap<_, _>>();
```
tailhook: You are creating an intermediate vector and a hashmap here. You don't need too. You may iterate over constant array like this: https://is.gd/oyNNsp (and yes, we usually avoid excessive memory allocations in rust)

lud4ik: fixed

src/builder/commands/pip.rs
```diff
@@ -313,31 +314,59 @@ impl BuildStep for Py3Install {
...
+               Result<(), VersionError> {
+
+    let path = try!(Path::new("/work")
+                       .join(fname)
+                       .canonicalize()
+                       .map_err(|e| VersionError::Io(e, fname.to_path_buf())));
+
+    let name = format!("{:?}", path);
+    if used.contains(&name[..]) {
+        return Ok(())
```
tailhook: I think we need error here. Does pip allow cyclic includes?

src/builder/commands/pip.rs
```diff
@@ -313,31 +314,59 @@ impl BuildStep for Py3Install {
...
+        let line = try!(line.map_err(|e| VersionError::Io(e, path.clone())));
+        let chunk = line[..].trim();
+        // Ignore empty lines and comments
+        if chunk.len() == 0 || chunk.starts_with("#") {
+            continue;
+        }
+        if let Some(req) = parse_req_filename(chunk) {
+            try!(version_req(hash,
+                             &fname.parent().unwrap().join(req),
+                             used));
```
tailhook: I think we need `continue` here. No need to hash line that does file include.

vagga.yaml
lud4ik: Deleted it because it's not related to the topic.
But is `prerequisites` and `-m` the same? There are examples for build as prerequisite for run.

src/builder/commands/pip.rs
lud4ik: for dependencies [yes](http://pip-python3.readthedocs.io/en/latest/reference/pip_install.html#installation-order)

lud4ik: pip 8.1.2:
`RuntimeError: maximum recursion depth exceeded while calling a Python object`

tailhook:
Merged in. Thanks!

PR #280, opened 2016-06-14
 #237 Got rid of `path_util::ToRelative` trait

anti-social:


src/builder/commands/generic.rs
```diff
@@ -84,14 +83,17 @@ fn find_cmd<P:AsRef<Path>>(ctx: &Context, cmd: P)
...
-                return Ok(path.join(cmd));
+            match path.strip_prefix("/") {
+                Ok(rel_path) => {
+                    if chroot.join(rel_path).join(cmd).exists() {
+                        return Ok(path.join(cmd));
+                    }
+                },
+                Err(_) => {
+                    warn!("All items in PATH must be absolute, not {}",
+                        path.display());
```
tailhook: better use `{:?}` instead of path display. The `path.display()` is legacy from 1.0 rust. 
(the reason is if path something weird like `\n` debug display will show it escaped, but `path.display()` prints a broken line)

tailhook:
Other than that, looks fine. That comment is not a regression so I'm merging in.

PR #281, opened 2016-06-15
Got rid of using `Path::display`

anti-social:


tailhook:
Merged path fix.

I think the mirror is fine, but it only contains recent versions of alpine. There are a  few more such mirrors. I don't think we should blacklist them. You may run tests with preferred mirror with `vagga -EALPINE_MIRROR=mirrorr`

Anyway, don't mix different changes in single pull request.

PR #276, opened 2016-06-13
Add elasticsearch example

alxpy:

PR #282, opened 2016-06-15
Add elasticsearch vagga example

tailhook:
Supersedes #276 

Any good ideas how to show case elasticsearch, like small python test script or interactive shell, or command-line utility or curl?

/cc @popravich , @alxpy, @anti-social


anti-social:
I prefer to install such software as elasticsearch from archive. In this case I can controll its exact version.

tailhook:
What about:
```
- !Install [elasticsearch=2.3.3]
```
I've also tried 2.2.0 and it works. So I'm pretty sure they keep old versions of packages in the repo.

PR #290, opened 2016-07-07
Added include-regex option for copy command

anti-social:


src/builder/commands/copy.rs
```diff
@@ -43,32 +90,31 @@ impl BuildStep for Copy {
         if src.starts_with("/work") {
             match symlink_metadata(src) {
                 Ok(ref meta) if meta.file_type().is_dir() => {
-                    let re = try!(Regex::new(&self.ignore_regex)
+                    let ignore_re = try!(self.compile_ignore_regex()
+                        .map_err(|e| VersionError::Regex(Box::new(e))));
```
tailhook: Why plain `try!` doesn't work here? Maybe add another `from()` clause?

src/builder/commands/copy.rs
```diff
@@ -43,32 +90,31 @@ impl BuildStep for Copy {
...
-                            strpath.map(|x| (fpath, x))
-                        }).collect::<Vec<_>>();
-                        all_entries.sort();
-                        for (fpath, name) in all_entries {
-                            hash.field("filename", name);
-                            try!(hash.file(&fpath,
-                                self.owner_uid, self.owner_gid)
-                                .map_err(|e| VersionError::Io(e, fpath)));
+                                generate_paths(strpath,
+                                    ignore_re.as_ref(), include_re.as_ref())
```
tailhook: You can return empty vector here instead of None, and use flat_map directly.
But I'm not sure if returning many one-element vectors is a good idea. Maybe there is another way that is also simple enough?

src/builder/commands/copy.rs
```diff
@@ -30,9 +32,54 @@ impl Copy {
...
+    match fpath.parent() {
+        Some(parent) => {
+            let mut cur_dir_path = PathBuf::new();
+            for dir_component in parent.components() {
+                match dir_component {
+                    Component::RootDir => cur_dir_path.push("/"),
+                    Component::Normal(dir_name) => cur_dir_path.push(dir_name),
+                    _ => unreachable!(),
+                }
+                result.push(cur_dir_path.clone());
```
tailhook: Wow! For each path `a/b/c/d` you **always** generate a list of allocated paths `a`, `a/b`, `a/b/c` ? I think it's too wasteful.

src/path_util.rs
```diff
@@ -40,3 +40,36 @@ impl Expand for PathBuf {
...
+
+impl<'a> Iterator for PathsIterator<'a> {
+    type Item = &'a Path;
+    fn next(&mut self) -> Option<&'a Path> {
+        match self.path {
+            Some(path) => {
+                if path == Path::new("") {
+                    None
+                } else {
+                    self.path = path.parent();
```
tailhook: I would make `path.parent().unwrap_or(Path::new(""))`, and get rid of `Option` in iterator. We treat `None` and `Path::new("")` the same anyway.

src/path_util.rs
```diff
@@ -40,3 +40,36 @@ impl Expand for PathBuf {
         }
     }
 }
+
+pub struct PathsIterator<'a> {
```
tailhook: Probably `ParentsIterator` is a better name.

src/builder/commands/copy.rs
```diff
@@ -30,9 +33,31 @@ impl Copy {
         .member("path", V::Directory::new().is_absolute(true))
         .member("ignore_regex", V::Scalar::new().default(
             r#"(^|/)\.(git|hg|svn|vagga)($|/)|~$|\.bak$|\.orig$|^#.*#$"#))
+        .member("include_regex", V::Scalar::new().optional())
         .member("owner_uid", V::Numeric::new().min(0).optional())
         .member("owner_gid", V::Numeric::new().min(0).optional())
     }
+
+    fn compile_ignore_regex(&self) -> Result<Option<Regex>, RegexError> {
```
tailhook: I would return a single structure e.g. `Filter`. which is then passed as a single structure to `check_path`.

tailhook:
Okay, just few small nitpicks this time. Overall looks solid.

src/path_util.rs
anti-social: Then `SelfAndParentsIterator` is the better name I think.

anti-social:
Is it worth to fold commits into one?

tailhook:
Looks good. Thanks!
(github allows to squash when pushing the merge button, so nevermind)

PR #293, opened 2016-07-08
Documentation for include-regex option

anti-social:
To disable `ignore-regex` we forced to use something like `^$` or `\A\z` which is not obvious. Should we disable `ignore-regex` initialized with empty string: `ignore-regex: ""`?


tailhook:
You need more text about including directories. If regex matches directory, only the folder, not contents will be included, right?

anti-social:
Updated.

PR #295, opened 2016-07-08
Improved Depends command: support directories and regular expression

anti-social:
Do not merge this PR!

As you can see two different containers `copy-with-include` and `depends-with-include` produce the same version hash. I think it would be necessary to include command name into hash but this will break hash evaluation again.

Also I can't choose name for `Depends`'s `path` option cause for `Copy` command `path` means destination path. `source` is a bad name too because it implies existance of destination path.


anti-social:
Adding destination `path` option into hash resolves problem for `!Copy` and `!Depends` commands.

anti-social:
Adding destination `path` option into hash resolves problem for `!Copy` and `!Depends` commands.

tailhook:
Looks fine. Is it ready for merge now?

anti-social:
I think yes.

tailhook:
Merged. Thanks!

PR #297, opened 2016-07-11
Fix php integration

livioribeiro:
Both Ubuntu xenial and Alpine 3.4 changed things for php.

Ubuntu added php7 and named its package just "php" with some modules not present in the main package compared to previous version (php-zip for instance).

Alpine changed the "php" package to "php5" in preparation for the upcoming "php7", which will probably be present in Alpine 3.5.

Ubuntu also added "hhvm" to the repository in xenail, so a test was added for that interpreter.

### DIFF
--- a/src/builder/commands/alpine.rs
+++ b/src/builder/commands/alpine.rs
@@ -21,6 +21,9 @@ static ALPINE_VERSION_REGEX: &'static str = r"^v\d+.\d+$";
 static MIRRORS: &'static str = include_str!("../../../alpine/MIRRORS.txt");
 
 
+const VERSION_WITH_PHP5: f32 = 3.4;
+
+
 // Build Steps
 #[derive(Debug)]
 pub struct Alpine(String);
@@ -70,7 +73,7 @@ impl Distribution for Distro {
         let mut to_install = vec!();
         let mut unsupp = vec!();
         for i in features.iter() {
-            if let Some(lst) = build_deps(*i) {
+            if let Some(lst) = self.build_deps(*i) {
                 for i in lst.into_iter() {
                     if !ctx.packages.contains(i) {
                         if ctx.build_deps.insert(i.to_string()) {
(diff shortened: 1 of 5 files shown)


src/builder/commands/alpine.rs
```diff
@@ -21,6 +21,9 @@ static ALPINE_VERSION_REGEX: &'static str = r"^v\d+.\d+$";
 static MIRRORS: &'static str = include_str!("../../../alpine/MIRRORS.txt");
 
 
+const VERSION_WITH_PHP5: f32 = 3.4;
```
tailhook: Well, keeping version as floating point is definitely wrong.

Either use a tuple `(3, 4)` or a string (`v3.4`), you can wrap latter to `config::version::Version` for correct version comparison (probably the wrapper might be moved out of config module)

src/builder/commands/ubuntu.rs
```diff
@@ -437,7 +437,7 @@ impl Distro {
             packages::Npm => Some(vec!()),
             // PHP
             packages::Php if self.has_php7() => {
-                Some(vec!("php-common", "php-cli"))
+                Some(vec!("php-common", "php-cli", "php-json", "php-zip"))
```
tailhook: Why do you need `php-json` and `php-zip`? Please add a comment into the source code.

src/builder/commands/composer.rs
```diff
@@ -316,7 +317,7 @@ fn find_conf_dirs() -> Result<Vec<PathBuf>, scan_dir::Error> {
     ScanDir::dirs().skip_symlinks(true).read("/vagga/root/etc", |iter| {
         iter.filter(|&(_, ref name)| name.starts_with("php"))
         .flat_map(|(ref entry, _)| {
-            ScanDir::dirs().read(entry.path(), |iter| {
+            ScanDir::dirs().walk(entry.path(), |iter| {
```
tailhook: Why do you need recursive walk?

livioribeiro: Ubuntu messed up with the config directory in xenial, now it's `/etc/php/7.0`. Without the recursive walk the config directory would not be found

tailhook:
Looks fine. But I'm still not sure about recursive walk.

Are there any incompatiblities introduced? Like different version of PHP installed on the same distribution, or something like that.

tailhook:
Looks good. Thanks!

PR #304, opened 2016-07-16
Do not respect mode when hashing Copy & Depends commands, umask option for Copy

anti-social:


src/builder/commands/copy.rs
```diff
@@ -104,6 +111,15 @@ impl BuildStep for Copy {
...
         }
         hash.field("path", self.path.to_str().unwrap());
+        if let Some(uid) = self.owner_uid {
+            hash.field("owner_uid", uid.to_string());
+        }
+        if let Some(gid) = self.owner_gid {
+            hash.field("owner_gid", gid.to_string());
+        }
+        if let Some(umask) = self.umask {
+            hash.field("umask", umask.to_string());
```
tailhook: Is there any need of hashing configuration parameters instead of hashing actual mode of the target file?

src/builder/commands/copy.rs
```diff
@@ -85,6 +91,7 @@ impl Copy {
         .member("include_regex", V::Scalar::new().optional())
         .member("owner_uid", V::Numeric::new().min(0).optional())
         .member("owner_gid", V::Numeric::new().min(0).optional())
+        .member("umask", V::Numeric::new().min(0).max(0o777).optional())
```
tailhook: The default of `umask` as None is unclear. Shouldn't we have default umask of `002` and a `preserve_permissions` flag instead? Or is it for backwards compatibility?

src/builder/commands/copy.rs
```diff
@@ -159,11 +185,33 @@ impl BuildStep for Copy {
...
+    let stat = try!(path.symlink_metadata()
+        .map_err(|e| StepError::Read(PathBuf::from(path), e)));
+    if stat.is_dir() {
+        let perms = Permissions::from_mode(DIR_MODE & !umask);
+        try!(set_permissions(path, perms)
+            .map_err(|e| StepError::Write(PathBuf::from(path), e)));
+    } else if stat.is_file() {
+        let orig_mode = stat.permissions().mode();
+        let mode = if orig_mode & EXE_CHECK_MASK > 0 {
+            DIR_MODE
```
tailhook: DIR_MODE? It would be more readable if you put `0o777` here, rather than a constant with a misleading name

src/builder/commands/copy.rs
```diff
@@ -159,11 +185,33 @@ impl BuildStep for Copy {
     }
 }
 
-fn hash_path(hash: &mut Digest, path: &Path,
-    filter: &Filter, owner_uid: Option<uid_t>, owner_gid: Option<gid_t>)
+fn reset_permissions(path: &Path, umask: u32)
+    -> Result<(), StepError>
+{
+    let stat = try!(path.symlink_metadata()
```
tailhook: I'm not sure this is a good idea:

1. We just created a file, we should know what permission bits to apply
2. Now we depend on both `umask` from `Copy` command **and** the environment (i.e.what if umask in Copy is 002 and umask in environment is 020)

src/builder/commands/copy.rs
```diff
@@ -206,6 +253,18 @@ fn hash_path(hash: &mut Digest, path: &Path,
     Ok(())
 }
 
+fn hash_file_and_exe_mode(hash: &mut Digest, path: &Path, stat: &Metadata)
+    -> Result<(), VersionError>
+{
+    if stat.is_file() {
+        let exe_mode = stat.permissions().mode() & EXE_CHECK_MASK;
+        hash.field("exe_mode", exe_mode.to_string());
```
tailhook: `hash.bool()` is more efficient

src/builder/commands/copy.rs
anti-social: Simplicity. Otherwise we should also calculate mode for each entry when hashing.

src/builder/commands/copy.rs
anti-social: At first I made default mode for copy command but then decided to keep backwards compatibility. Also I didn't like idea to add yet another option.

src/builder/commands/copy.rs
tailhook: A thought experiment: Where will you put a notice that non-specifing umask makes `image-cache-url` partially useless?

I mean people need always specify umask, or otherwise they will complain that `image-cache-url` doesn't work, and most of the time it would be hard to diagnose the problem.

src/builder/commands/copy.rs
tailhook: A thought experiment: Where will you put a notice that non-specifing umask makes `image-cache-url` partially useless?

I mean people need always specify umask, or otherwise they will complain that `image-cache-url` doesn't work, and most of the time it would be hard to diagnose the problem.

src/builder/commands/copy.rs
anti-social: 1. In this case I should patch `file_util::copy` & `file_util::shallow_copy` functions. Possible is it worth to add `mode` optional parameter to these functions?
2. Only `umask` from `Copy` should be applied. We only take executable bit from created file.

anti-social:
What default umask should we chose for `Copy` command, 002 or 022?

src/builder/commands/copy.rs
```diff
@@ -186,27 +180,24 @@ impl BuildStep for Copy {
     }
 }
 
-fn reset_permissions(path: &Path, umask: u32)
-    -> Result<(), StepError>
+fn calc_mode(path: &Path, umask: u32)
```
tailhook: We already have metadata in the loop above. Just pass it here instead of a Path.

src/builder/commands/copy.rs
```diff
@@ -104,6 +115,13 @@ impl BuildStep for Copy {
             // inside the container which is ugly
         }
         hash.field("path", self.path.to_str().unwrap());
+        if let Some(uid) = self.owner_uid {
+            hash.field("owner_uid", uid.to_string());
+        }
+        if let Some(gid) = self.owner_gid {
+            hash.field("owner_gid", gid.to_string());
+        }
+        hash.field("umask", format!("{:o}", self.umask));
```
tailhook: I don't think octal formatting is useful for hashing. I think we will get rid of `.to_string()` in `hash.field` and having to `format!` will make that optimization harder.

tailhook:
Okay, sorry `calc_mode` optimization was useless in the way it currently is. And still `shallow_copy` and `calc_mode` combination require two `symlink_metadata` calls. I still think it's possible to optimize.

Another observation is that `Digest::file` looks like to depend on `Copy` too much. I think it must be removed from `digest.rs` and put as a function into `copy.rs`.

And I'm still not sure about hashing. Talking about efficiency: we already hash "executable" bit, hashing all the bits will not make any noticeable performance difference. So we are talking only about maintenance and UX issues. 

Generally, vagga tries to ignore unimportant facts. For example, renaming `package.json` or `requirements.txt` doesn't influence container hash. So it's tempting do the same here.

Will it require much more code? At a glance, it looks like you just need to change a `hash_file_and_exe_mode` function.

anti-social:
> And I'm still not sure about hashing. Talking about efficiency: we already hash "executable" bit, hashing all the bits will not make any noticeable performance difference. So we are talking only about maintenance and UX issues. 

`Depends` command doesn't have `umask` option thus it doesn't depend on file mode only executable bit.

> Generally, vagga tries to ignore unimportant facts. For example, renaming package.json or requirements.txt doesn't influence container hash. So it's tempting do the same here.

It is not clear for me. Do you want to get rid of hashing `umask` but hash file permissions instead?

tailhook:
So we are agreed that for paths in inside the `/work` owner_uid/owner_gid/umask should be hashed as he part of the file attributes rather than as field in the `Copy` structure. Right?

anti-social:
Definitely. Try to finish this PR in the near future.

src/builder/commands/copy.rs
```diff
@@ -94,16 +143,38 @@ impl BuildStep for Copy {
...
+            hash.field("path", self.path.to_str().unwrap());
         } else {
             // We don't version the files outside of the /work because
             // we believe they are result of the commands run above
             //
             // And we need already built container to version the files
             // inside the container which is ugly
+            hash.field("source", src.to_str().unwrap());
+            hash.field("path", self.path.to_str().unwrap());
+            if !self.preserve_permissions {
```
anti-social: Should we hash `self.preserve_permissions` flag? If `preserve_permissions` is `true` we will hash `self.umask` so hashes will be different.

tailhook:
Merged. Congrats!

And can you please update the docs?

Issue #264, opened 2016-05-27
Add a warning if `vagga.yml` is present in the directory hierarcy


zemlanin:
🙍

tailhook:
Well, I don't think it should. Mostly because if we support both `yml` and `yaml` there is an ambiguity of which is chosen when both are present.

I think the times when more than three character extension was a problem have already passed.

tailhook:
By the way, if it is just common case for mistakes you're doing, we can add a warning if there is `vagga.yml` but not `vagga.yaml`. Is it something you appreciate?

zemlanin:
It totally is. I guess, warning when there are both files can be helpful too

tailhook:
Okay, I'm reopening the issue and marking it "easy". If no-one provides a pull request I'll fix it later.

PR #307, opened 2016-07-18
#264 Added warning when vagga.yml is present but not vagga.yaml

anti-social:


src/config/config.rs
```diff
@@ -41,25 +42,45 @@ pub fn config_validator<'a>() -> V::Structure<'a> {
...
+                    possibly it is a typo. \
+                    Correct configuration file name is vagga.yaml",
+                dir).ok();
+        }
+    }
+    
     let mut dir = work_dir.clone();
     loop {
         let fname = dir.join(".vagga/vagga.yaml");
         if fname.exists() {
```
tailhook: Warning should be printed even if `vagga.yaml` is present too.

src/config/config.rs
```diff
@@ -41,25 +42,45 @@ pub fn config_validator<'a>() -> V::Structure<'a> {
         command_validator()))
 }
 
-fn find_config_path(work_dir: &PathBuf) -> Option<(PathBuf, PathBuf)> {
+fn find_config_path(work_dir: &PathBuf, show_warnings: bool)
+    -> Option<(PathBuf, PathBuf)>
+{
+    fn maybe_print_warning(dir: &Path) {
```
tailhook: Why this function is inside? I don't see clear reason keeping it there.

src/config/config.rs
```diff
@@ -41,25 +42,45 @@ pub fn config_validator<'a>() -> V::Structure<'a> {
...
+        if dir.join("vagga.yml").exists() {
+            writeln!(&mut stderr(), "There is vagga.yml file in the {:?}, \
```
tailhook: We usually to warnings with a `warn!` macro. The positive outcome of using `warn!` macro is that user can silence all warnings or some of them using `RUST_LOG`

src/config/config.rs
anti-social: > we can add a warning if there is vagga.yml but not vagga.yaml

I think this is reasonable to print warning if only incorrect file name is present.

src/config/config.rs
anti-social: Cause it is needed only to get rid of copy-pasting message.

src/config/config.rs
tailhook: So what to do if both are present? Isn't it confusing when you fix `vagga.yml` nothing changes?

Anyway, current fix looks fine.

PR #308, opened 2016-07-18
Fixed path precedence for _run command

anti-social:


src/wrapper/run.rs
```diff
@@ -70,16 +71,8 @@ pub fn run_command_cmd(wrapper: &Wrapper, cmdline: Vec<String>)
...
-        let paths = [
-            "/bin",
-            "/usr/bin",
-            "/usr/local/bin",
-            "/sbin",
-            "/usr/sbin",
-            "/usr/local/sbin",
-        ];
+    if !command.contains("/") {
+        let paths: Vec<_> = DEFAULT_PATH.split(':').collect();
```
tailhook: You don't need `.collect()` here. Just iterate over split parts.

tailhook:
Is there anything that have fixed by this change?

src/wrapper/run.rs
anti-social: There is usage of `paths` below:
```
         if !cpath.is_absolute() {
             return Err(format!("Command {:?} not found in {:?}",
                 cpath, &paths));
}
```

Or we can use `DEFAULT_PATH` instead of array.

tailhook: Yes, printing colon-separated PATH should be fine.

anti-social:
I updated ubuntu version for some gem_bundler tests and they failed.

tailhook:
> I updated ubuntu version for some gem_bundler tests and they failed.

There are two bundlers in ubuntu?

anti-social:
The problem binary is rake. Think one is installed as dependency with ruby-dev package inside /usr/bin and another inside /usr/local/bin when updating gem.

tailhook:
Hopefully this will not break many things.

PR #313, opened 2016-07-26
Added container option for Snapshot volume

anti-social:


src/wrapper/setup.rs
```diff
@@ -441,7 +442,18 @@ pub fn setup_filesystem(setup_info: &SetupInfo, container_ver: &str)
                 try!(remount_ro(&dest));
             }
             &V::Snapshot(ref info) => {
-                try!(make_snapshot(&dest, info));
+                let ref src = match info.container {
+                    Some(ref src_container) => {
+                        let container_ver = try!(version_from_symlink(
```
tailhook: We are trying not to use `version_from_symlink` if possible. I.e. use computed hash rather than a symlink.

tailhook: Okay, given that we use the same for `Container` volume, it's probably unfair requirement :)

tailhook:
Merged. Thanks!

PR #318, opened 2016-07-27
Implemented data-dirs container option

anti-social:


src/builder/context.rs
```diff
@@ -19,6 +19,9 @@ use super::timer;
 use file_util::create_dir;
 use process_util::PROXY_ENV_VARS;
 
+pub const PROTECTED_DIRS: &'static [&'static str] = &[
+    "dev", "proc", "run", "sys", "tmp", "work",
```
tailhook: Are there any technical reasons to protect them? I.e. probably this kind of "shrinked" container won't be used for running commands anyway, so these mount ponts aren't going to be useful

src/version/version.rs
```diff
@@ -26,6 +28,13 @@ fn all(container: &Container, cfg: &Config)
         try!(b.hash(&cfg, &mut hash).map_err(|e| (format!("{:?}", b), e)));
     }
 
+    if !container.remove_all_except.is_empty() {
+        let remove_all_strings = container.remove_all_except.iter()
+            .map(|p| p.as_os_str().as_bytes())
+            .collect::<Vec<_>>();
+        hash.sequence("remove_all_except", &remove_all_strings);
```
tailhook: Make a function that accepts iterable. Or update `hash.sequence` if it doesn't break anything.

src/builder/guard.rs
```diff
@@ -100,6 +104,31 @@ impl<'a> Guard<'a> {
                 "Error creating dir: {err}");
         }
 
+        if !self.ctx.container_config.remove_all_except.is_empty() {
+            let root = Path::new("/vagga/root");
+            let exclude_paths = PROTECTED_DIRS.iter()
```
tailhook: Why `exclude_paths` is a `Vec<PathBuf>` if it could be `Vec<&Path>`? Or just `Iterator<&Path>` (actually a `Chain..` but that's irrelevant)

src/builder/guard.rs
```diff
@@ -109,3 +138,37 @@ impl<'a> Guard<'a> {
         return Ok(());
     }
 }
+
```
tailhook: I need a comment what `true`/`false` in keep_rel_path means. Or make two sets with more specific names.

src/builder/guard.rs
anti-social: > Why exclude_paths is a Vec<PathBuf> if it could be Vec<&Path>? Or just Iterator<&Path> (actually a Chain.. but that's irrelevant)

With iterator there was a lifetime issue that I couldn't resolve.

src/builder/guard.rs
```diff
@@ -100,6 +104,31 @@ impl<'a> Guard<'a> {
...
+                    .map(|p| PathBuf::from(p.strip_prefix("/").unwrap())))
+                .collect::<Vec<_>>();
+            let mut keep_rel_paths = HashMap::new();
+            for exclude_path in &exclude_paths {
+                for (i, p) in exclude_path
+                    .iter_self_and_parents().enumerate()
+                {
+                    if i == 0 {
+                        keep_rel_paths.insert(p, true);
+                    } else if !keep_rel_paths.contains_key(p) {
```
tailhook: Shouldn't the duplicate key here produce a warning? (in both cases of the if statement). I mean it's something like `[/var, /var/lib]`, which is mistake most of the time, unless you're trying to merge multiple lists which i don't think is a common case.

src/builder/guard.rs
```diff
@@ -100,6 +104,31 @@ impl<'a> Guard<'a> {
...
+                    .iter()
+                    // We validate exclude paths as absolute
+                    .map(|p| PathBuf::from(p.strip_prefix("/").unwrap())))
+                .collect::<Vec<_>>();
+            let mut keep_rel_paths = HashMap::new();
+            for exclude_path in &exclude_paths {
+                for (i, p) in exclude_path
+                    .iter_self_and_parents().enumerate()
+                {
+                    if i == 0 {
```
tailhook: I think it's better to `skip(1) ` from an iterator rather than enumerating and checking `i == 0`

src/builder/context.rs
anti-social: Thought about this but we don't know how somebody will use such containers. So I decided that several empty directories are not a big deal.

tailhook:
Besides few comments in the code I have two thoughts:

1. The `remove-all-...` thing is too imperative. The name would be fine inside the `setup`. But here it's unclear why it will remove something and when. But I'm not sure about a good name. Maybe `keep-only-dirs`. Or maybe some wordplay on "data container"
2. I think we should add a warning when running a command inside such container. Have you tried that? It will probably crash with a weird error message. Also, I think we should remove it from completion of the `vagga _run` but that's not so important.

anti-social:
Possibly something relatid with "data container" would be a good name. It could explain missing all of the service directories.

anti-social:
`data-only`, `data-only-container`?

tailhook:
For `data-only` I would expect `true`

So maybe `data-dirs` ?

src/builder/guard.rs
```diff
@@ -8,8 +9,10 @@ use builder::commands::{composer, gem, npm, pip};
 use builder::packages;
 use build_step::BuildStep;
 use container::util::clean_dir;
-use container::mount::{unmount, mount_system_dirs, mount_proc};
+use container::mount::{unmount, mount_system_dirs, unmount_system_dirs,
+    mount_proc};
```
tailhook: We don't wrap `use` lines. We use two separate `use` statements instead.

src/builder/guard.rs
```diff
@@ -109,3 +143,41 @@ impl<'a> Guard<'a> {
...
+                        path=path);
+                }
+            },
+        }
+    }
+    Ok(())
+}
+
+fn warn_duplicate_data_dir(path: &Path) {
+    warn!("{:?} is already marked as data directory", Path::new("/").join(path));
```
tailhook: The warning is a little bit misleading:
```
[/var/log, /var]
->  "/var" is already marked as data directory
```
Well, it's not.

You can use a boolean value from `keep_rel_paths` to refine warning. Or you can describe both cases, like ".. is either a prefix of other directory or already contained in another prefix"

src/wrapper/util.rs
```diff
@@ -28,3 +31,11 @@ pub fn find_cmd(cmd: &str, env: &BTreeMap<String, String>)
         }
     }
 }
+
+pub fn check_data_container(container_config: &Container) {
```
tailhook: `check_data_container` -> `warn_if_data_container`

Current name implies some check if data container has appropriate settings, rather than what it does

anti-social:
Updated.

tailhook:
I think `data-dirs` is fine for now, but we can rename it for 0.7 if that turns out to be a nightmare.

Add the documentation please!

PR #321, opened 2016-08-03
AlpineRepo and Repo commands

anti-social:
Doubt about naming of `alias` option for `AlpineRepo` command. In the [Alpine Wiki](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management) they call it as `pin` or `tag`.

### DIFF
--- a/docs/build_steps.rst
+++ b/docs/build_steps.rst
@@ -229,7 +229,7 @@ Alpine Commands
        url: http://nl.alpinelinux.org/alpine/
        branch: edge
        repo: testing
-       alias: testing
+       tag: testing
      - !Install [app@testing]
 
    Options:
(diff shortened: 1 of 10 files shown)


docs/build_steps.rst
```diff
@@ -215,6 +221,36 @@ Alpine Commands
...
+     Branch of the repository. For example ``v3.4``, ``edge``. Default is
+     the version of the current alpine distribution.
+
+   repo
+     Repository to fetch packages from. For example ``main``, ``community``,
+     ``testing``. **Required**.
+
+   alias
+     Alias or "pin" for this repository. Alpine package manager will now
+     by default only use the untagged repositories. Adding an alias to
```
tailhook: "Alias or pin" and later "untagged repositories", it's unclear what "untagged" means

docs/build_steps.rst
```diff
@@ -215,6 +221,36 @@ Alpine Commands
...
+     specific package will prefer the repository with that alias.
```
tailhook: How to add an alias?

tailhook:
I think `tag` would be fine. The word "pin" is only used as a verb in the wiki, AFAICS.

tailhook:
Except documentation refinements and `tag` it's good to go.

tailhook:
Merged. Thanks!

PR #322, opened 2016-08-03
Updated docs: data directories and snapshot volume options

anti-social:


tailhook:
I'm merging this in. As it's better than nothing. But `data-dirs` must have a better example...

Issue #284, opened 2016-06-17
!Remove does not remove symbolic links


Orhideous:
Minimal working `vagga.yaml`
```yaml
containers:
  test:
    setup:
      - !Alpine v3.3
      - !Sh |
          mkdir -p /opt/test_dir
          ln -s /opt/test_dir -T /opt/test_symlink

  inherited:
    setup:
      - !Container test
      - !Remove /opt/test_symlink


commands:
  check: !Command
    container: inherited
    run: ls -hal /opt
```

`vagga check `
```
total 12
drwxrwxr-x    3 root     root        4.0K Jun 17 11:59 .
drwxr-xr-x   19 root     root        4.0K Jun 17 11:59 ..
drwxrwxr-x    2 root     root        4.0K Jun 17 11:59 test_dir
lrwxrwxrwx    1 root     root          13 Jun 17 11:59 test_symlink -> /opt/test_dir
```
As we can see,  `test_symlink` wasn't removed, but it should.

tailhook:
Should be fixed by #326

PR #326, opened 2016-08-05
Remove command should be able to remove symlinks #284

anti-social:


src/builder/guard.rs
```diff
@@ -89,18 +89,17 @@ impl<'a> Guard<'a> {
...
         }
 
         for dir in self.ctx.empty_dirs.iter() {
             try!(clean_dir(&base.join(dir), false));
         }
 
         for dir in self.ctx.ensure_dirs.iter() {
-            let fulldir = base.join(dir);
-            try_msg!(create_dir(&fulldir, true),
+            try_msg!(dirs::ensure(&Path::new("/").join(dir)),
```
tailhook: Sorry, is it correct to change `base` (which is `/vagga/root`) to `/` ?

anti-social: Yes, it is correct. `/vagga/root` will prepend inside `ensure` function.

tailhook:
Merged. Thanks!

PR #335, opened 2016-08-27
Fixed command to run psql in postgresql example

brabadu:
Had to specify `$PG_PORT` in psql call as it tries 5432 port by default, while our env set to 5433


tailhook:
Merged. Thanks!

PR #340, opened 2016-09-06
New postgres example

tailhook:
This example employs latest features in `!Persistent`.

Changes:

1. Sets fixed user id 200 for postgres, because volume usually outlives a container
2. Now uses persistent volume instead of tmpfs
3. Cleaner initialization command that uses `user-id` instead of `su` (also works on all host systems)
4. Adds example of how to prefill the database with data
5. Removes useless environment variables

/cc @anti-social, @popravich, @Orhideous, @brabadu

### DIFF
--- a/examples/postgres/vagga.yaml
+++ b/examples/postgres/vagga.yaml
@@ -6,31 +6,76 @@ containers:
   ubuntu:
     setup:
     - !Ubuntu xenial
-    - !Install
-      - postgresql-9.5
+    # Use fixed user id and group id for postgres, because in some cases
+    # we may need to upgrade (rebuild) a postgres container, but keep the data
+    # on a `!Persistent` volume still usable. User ids in ubuntu packages are
+    # not guaranteed to be same on every installation.
+    #
+    # The command-lines are from the postgres-common package except
+    # added --uid 200 --gid 200
+    - !Sh |
+        addgroup --system --gid 200 postgres
+        adduser --uid 200 --system --home /data --no-create-home \
+            --shell /bin/bash --group --gecos "PostgreSQL administrator" \
+            postgres
+    - !Install [postgresql-9.5]
     - !EnsureDir /data
     environ:
       PG_PORT: 5433   # Port of host to use
       PG_DB: vagga-test
-      PG_USER: vagga
-      PG_PASSWORD: vagga
       PGDATA: /data
       PG_BIN: /usr/lib/postgresql/9.5/bin
     volumes:
-      /data: !Tmpfs
-        size: 100M
-        mode: 0o700
+      /data: !Persistent
+        name: postgres
+        owner-uid: 200
+        owner-gid: 200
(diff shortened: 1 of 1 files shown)


examples/postgres/vagga.yaml
```diff
@@ -6,31 +6,76 @@ containers:
...
+    description: Init postgres database
+    container: ubuntu
+    user-id: 200
+    group-id: 200
+    run: |
+      set -ex
+      ls -la /data
+      $PG_BIN/pg_ctl initdb
+      $PG_BIN/pg_ctl -w -o '-F --port=$PG_PORT -k /tmp' start
+      $PG_BIN/psql -h 127.0.0.1 -p $PG_PORT \
```
Orhideous: Maybe, with [createuser](https://www.postgresql.org/docs/9.5/static/app-createuser.html) this will look prettier?
`$PG_BIN/createuser -h 127.0.0.1 -p $PG_PORT vagga`

examples/postgres/vagga.yaml
```diff
@@ -6,31 +6,76 @@ containers:
...
+        -c "CREATE USER vagga WITH PASSWORD 'vagga';"
+      $PG_BIN/createdb -h 127.0.0.1 -p $PG_PORT $PG_DB -O vagga
+
+      # init schema usually schema shouldn't be inline here, but contained
+      # in a separate file
+      psql postgres://vagga:vagga@127.0.0.1:$PG_PORT/$PG_DB <<ENDSQL
+      CREATE TABLE random_stuff (x TEXT);
+      ENDSQL
+
+      pkill postgres
```
Orhideous: `$PG_BIN/pg_ctl stop` ?

Orhideous:
Looks much better afterall :+1:

examples/postgres/vagga.yaml
```diff
@@ -6,31 +6,76 @@ containers:
...
-      - postgresql-9.5
+    # Use fixed user id and group id for postgres, because in some cases
+    # we may need to upgrade (rebuild) a postgres container, but keep the data
+    # on a `!Persistent` volume still usable. User ids in ubuntu packages are
+    # not guaranteed to be same on every installation.
+    #
+    # The command-lines are from the postgres-common package except
+    # added --uid 200 --gid 200
+    - !Sh |
+        addgroup --system --gid 200 postgres
```
Orhideous: Maybe `adduser`/`addgroup` commands should became `!AddUser` in vagga, shouldn't it?

tailhook: Basically, we add commands for one of the three reasons:

1. If expressing it as a shell command is too long and hard to remember
2. If we can make much better versioning of this thing if it's written as a built-in command
3. If it's impossible to do using command-line

I think neither one is applied to this command. These ones are well known and well documented. We aren't going to reimplement every single command-line utility :)

PR #348, opened 2016-09-18
Fixed Tar command when unpacking archive with missing intermediate directories

anti-social:


src/file_util.rs
```diff
@@ -33,29 +33,82 @@ pub fn read_visible_entries(dir: &Path) -> Result<Vec<PathBuf>, Error> {
...
+        self
+    }
+
+    pub fn create(&self) -> Result<(), Error> {
+        create_dir(self.path.as_ref(), self.recursive,
+            self.mode, self.uid, self.gid)
+    }
+}
+
+pub fn create_dir(path: &Path, recursive: bool,
```
tailhook: If you don't keep this function for backward compatibility (i.e. you change arguments anyway). then don't make it public anymore. If it's not public just make it a method and don't "unpack" structure items manually.

src/file_util.rs
```diff
@@ -33,29 +33,82 @@ pub fn read_visible_entries(dir: &Path) -> Result<Vec<PathBuf>, Error> {
...
+    if uid.is_some() || gid.is_some() {
+        let uid = if let Some(uid) = uid {
+            uid
+        } else {
+            try!(path.symlink_metadata()).uid()
+        };
+        let gid = if let Some(gid) = gid {
+            gid
+        } else {
+            try!(path.symlink_metadata()).gid()
```
tailhook: You make a "stat()" call twice in a row here. In some cases, like unpacking a large archive it can make a large difference.

anti-social: Only once cause we enter in this block only if `uid.is_some() || gid.is_some()`.

tailhook: Ah, yes I overlooked that.

anti-social:
`.recursive_with_permissions(true)` method will be ok?

anti-social:
Think `.propagate_permissions(true)` is better because `.recursive_with_permissions(false)` is ambigious.

anti-social:
> On the other hand, if we have a !Snapshot which has owner enforced in configuration file we must create intermediate directories with that owner.

As I can see we don't create mount point directories. So at this moment propagating permissions is useless.

src/file_util.rs
anti-social: If I make it a method then it will be compiled for every generic `P: AsRef<Path>`.

anti-social:
Updated.

tailhook:
Merged. Thanks!

PR #351, opened 2016-09-20
Fix for multiline command descriptions indentation

popravich:
1. Having command description as follows will print descriptions unindented (second+ lines of it):
  ```yaml
  command:
    run: !Command
      description: |
        Some note on common options to use:
        vagga run
        vagga run -w
  ```
  ```bash
  $ vagga
  Available commands:
      run             Some note on common options to use:
  vagga run
  vagga run -w
  ```
  
  The PR fixes indent.

2. Also `vagga _list -A` doesn't show *all* commands -- only builtin so PR addresses this too.

### DIFF
--- a/src/launcher/list.rs
+++ b/src/launcher/list.rs
@@ -53,7 +53,14 @@ pub fn print_list(config: &Config, mut args: Vec<String>)
                         }
                         out.write_all(b" ").ok();
                     }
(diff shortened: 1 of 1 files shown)


tailhook:
Merged. Thanks!

This is good enough for now. In perspective, I think we should rewrap description entirely.

PR #354, opened 2016-09-26
'vagga' should not capture tty by default

anti-social:


src/process_util.rs
```diff
@@ -135,8 +153,11 @@ pub fn run_and_wait(cmd: &mut Command)
...
                 }
-                try!(tty_guard.check().map_err(|e|
-                    format!("Error handling tty: {}", e)));
+                if take_tty {
+                    try!(tty_guard.check().map_err(|e|
+                        format!("Error handling tty: {}", e)));
+
+                }
             }
             _ => unreachable!(),
```
tailhook: Still crashes on SIGSTTIN, SIGTTOU, right?

I mean all trapped singnals should be handled somehow.

src/process_util.rs
```diff
@@ -79,15 +79,15 @@ pub fn capture_fd3_status(mut cmd: Command)
     Ok((status, buf))
 }
 
-pub fn run_and_wait(cmd: &mut Command)
+pub fn run_and_wait(cmd: &mut Command, take_tty: bool)
```
tailhook: Can you explain in docstring here when we need take_tty of true or false? It's not obvious at all. And I think it might hide some bug too.

src/process_util.rs
anti-social: Agree. Should we ignore those signals or propagate `SIGTSTP` to children and send `SIGSTOP` self? (think we should stop all processes)

Now I can't reproduce trapping vagga `SIGTTIN` or `SIGTTOU` signals (only manually via `kill` command).

It is very interesting that vagga launcher received `SIGTTIN` (I rather expected `SIGTTOU`) when stdout was redirected into `less`:

```
$ ps -g 7135 -H -o pid,pgrp,sid,tty,stat,cmd
  PID  PGRP   SID TT       STAT CMD
 7135  7135  7135 pts/20   Ss   -zsh
 9495  9495  7135 pts/20   S      ../../vagga _run py2-alpine python -c  import time time.sleep(2) for i in range(1000000):     print i 
 9504  9504  7135 pts/20   S+       vagga_wrapper --root py2-alpine.202203b7 _run py2-alpine python -c  import time time.sleep(2) for i in rang
 9507  9507  7135 pts/20   S          /usr/bin/python -c  import time time.sleep(2) for i in range(1000000):     print i 
 9496  9495  7135 pts/20   S      less
```

Think this was due to taking tty by `vagga_wrapper` but still don't understand why did launcher receive `SIGTTIN`?

tailhook: > SIGTTOU is only generated for an attempt to write to the terminal if the TOSTOP output mode is set; see Output Modes.

This is why SIGTTOU is not generated

> It is very interesting that vagga launcher received SIGTTIN

Probably `less` tried to read and whole process group was sent SIGTTIN to

> Should we ignore those signals or propagate SIGTSTP to children and send SIGSTOP self?

It looks like we should stop all children:
```
> bash | less &
[1] 12057 12058
[1]  + suspended (tty output)  bash | less
```

anti-social: > Probably `less` tried to read

Oh yeah, `less` needs TTY, not only reading from pipe.

src/launcher/supervisor.rs
```diff
@@ -356,30 +358,25 @@ pub fn run(sup: &SuperviseInfo, args: Args, data: Data,
         // Normal loop
         'signal_loop: for signal in trap.by_ref() {
             match signal {
-                SIGINT => {
-                    // SIGINT is usually a Ctrl+C, if we trap it here
-                    // child process hasn't controlling terminal,
+                SIGINT|SIGQUIT => {
+                    // SIGINT is usually a Ctrl+C and SIGQUIT is a Ctrl+\,
+                    // if we trap it here child process hasn't controlling terminal,
                     // so we send the signal to the child process
```
tailhook: Please move comment to a docstring of the function

src/launcher/supervisor.rs
```diff
@@ -356,30 +358,25 @@ pub fn run(sup: &SuperviseInfo, args: Args, data: Data,
...
-                SIGTERM|SIGQUIT => {
+                SIGTSTP|SIGTTIN|SIGTTOU => {
+                    process_tty_signal(signal, &children);
+                }
+                SIGCONT => {
+                    process_sigcont(&children);
+                }
+                SIGTERM => {
                     // SIGTERM is usually sent to a specific process so we
                     // forward it to children
```
tailhook: Please move comment to a docstring of the function

src/tty_util.rs
```diff
@@ -41,13 +41,17 @@ impl TtyGuard {
     pub fn check(&mut self) -> Result<(), io::Error> {
         let &mut TtyGuard { ref tty, my_pgrp, .. } = self;
         tty.as_ref().map_or(Ok(()), |f| {
-            if try!(unsafe { get_group(f.as_raw_fd()) }) == 0 {
-                try!(unsafe { give_tty_to(f.as_raw_fd(), my_pgrp) });
+            let tty_owner_grp = try!(unsafe { get_group(f.as_raw_fd()) });
+            if tty_owner_grp != 0 {
+                let kill_res = unsafe { kill(tty_owner_grp, 0) };
+                if kill_res < 0 && io::Error::last_os_error().raw_os_error().unwrap() == ESRCH {
```
tailhook: There is `errno()` function in libc and nix. Wrapping and unwrapping looks weird.

BTW why don't you use `kill` from `nix` ? It returns an error that you can match against.

anti-social: Cannot find a way to pass into nix's `kill` 0 signum.

```

tailhook:
Merged. Thanks!

PR #357, opened 2016-10-05
'vagga _version_hash' shouldn't panic when container is not found

anti-social:


tailhook:
Have you verified that error message is good enough?

anti-social:
The message is:
```
Container "unknown" not found
```

It is short and clear I think.

tailhook:
Okay. Merged!

PR #368, opened 2016-11-28
Fix postgres in tutorials

### LINKED ISSUE #364: Fix postgres usage in tutorial
There are two issues:

1. You can now use user instead of `su` making posgres example [much simpler](https://github.com/tailhook/vagga/blob/master/examples/postgres/vagga.yaml)

2. Postgres doesn't shutdown properly if run with pg_ctl:
```
waiting for server to start....LOG:  database system was shut down at 2016-11-14 21:49:18 UTC
LOG:  MultiXact member wraparound protections are now enabled
LOG:  autovacuum launcher started
LOG:  database system is ready to accept connections
 done
server started
^CLOG:  received fast shutdown request
LOG:  aborting any active transactions
```
Correct shutdown is:
```
LOG:  received fast shutdown request
LOG:  aborting any active transactions
LOG:  autovacuum launcher shutting down
LOG:  shutting down
LOG:  database system is shut down
```
It can be achieved by using trap:
```yaml
    run: |
      trap "$PG_BIN/pg_ctl -w stop; trap - INT; kill -INT $$" INT
      $PG_BIN/pg_ctl -w -o '-F --port=$PG_PORT -k /tmp' start
      sleep infinity
```
Or running postgres with `exec` as in example above.

/cc @livioribeiro if you have time to look at it

livioribeiro:
Closes #364 

Also, Laravel/PHP tutorial was changed to use Postgres instead of MySQL because Postgres was a bit easier to work with.


tailhook:
Great job. Thanks!

Any specific issues with mysql? I'm going to add mysql into examples too because it's another thing a lot of people are interested in.

livioribeiro:
I just tried to initialize the database volume the same way we did with PostgreSQL, but I could not get it to work, I was probably missing something.

PR #373, opened 2016-12-08
New hashes

### LINKED ISSUE #366: Invalid calculations of uidmap
Here is relevant panic.
```$ vagga _clean --unused
thread '<main>' panicked at 'arithmetic operation overflowed', src/config/range.rs:52
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

This is a bug in range calculations. Fixing it in most straightforward and sensible way means changing hashes of all containers.

We can either patch only this and keep technical debt by hardcoding `-1` in hash function. Or fix it in most sensible way and require everyone to rebuild (all hashes will change). If we choose latter we might take a step further to clean up technical debt from all command hash functions.

/cc @anti-social

tailhook:
As an excuse to fix #366 while breaking hashes I'm going to refactor all the hashing code and implement some debug log for the hashed data.

ping @anti-social


anti-social:
Are you going to add build step names into hash?

tailhook:
> Are you going to add build step names into hash?

Yes. Just got no time to do it overnight :)

tailhook:
Done. @anti-social any other suggestions?

anti-social:
Really great job!

Issue #358, opened 2016-10-11
`remount_ro` should obey `nosuid,nodev`


tailhook:
In current setup when we are trying to remount to readonly the filesystem set up with `nosuid,nodev` we are getting a `Permission denied` error.

Same issue is probably with running vagga on tmpfs that we observed earlier.

roman901:
I dont think that it can be fixed. Vagga suggests that suid will be modifyed, as I understanded

tailhook:
I'm not sure. But presumably, we must read flags from the initial mountpoint and pass them when doing `mount(...MS_REMOUNT,MS_RDONLY..)`. (Otherwise, kernel thinks we are trying to reset them, i.e. get rid of nosuid, nodev)

PR #374, opened 2016-12-08
Keep mount point options when remounting readonly #358

anti-social:


tailhook:
Don't you want to add `BindMount().readonly(true)` ?

PR #377, opened 2016-12-10
Fixed running faketime when building container

anti-social:


src/builder/guard.rs
```diff
@@ -46,6 +46,7 @@ impl<'a> Guard<'a> {
 
     pub fn start(&mut self) -> Result<(), String> {
         mount_system_dirs()?;
+        mount_run(&Path::new("/vagga/root/run"))?;
```
tailhook: Any particular reason of why mount_run is not in mount_system_dirs?

src/builder/commands/ubuntu.rs
```diff
@@ -216,31 +222,18 @@ impl Distribution for Distro {
...
-                    if m.len() > 0 {
-                        m.push(':');
-                        m.push_str(params.preload);
-                    } else {
-                        m.push_str(params.preload);
-                    }
-                }
+            if let Some(ref params) = eatmy {
+                apt_get_install(ctx, &[params.package], &EatMyData::No)?;
+                self.eatmydata = EatMyData::Preload(params.preload.to_string());
```
tailhook: Preload is currently a always is a static string, why do you clone it here, and then in each command invocation?

src/builder/guard.rs
anti-social: In order to not mount `/run` twice when somebody sets volume into `/run`. See https://github.com/tailhook/vagga/pull/377/commits/aa13159c8266885555e28f17115dc56b62cf9bb1#diff-2713dfcee40c652f64f1994c5200e045L408

PR #387, opened 2017-01-19
Fix php in alpine 3.5

### LINKED ISSUE #385: On Alpine v3.5, install php7 from community or php5 from main?
Alpine v3.5 was launched with php7 available but only in community repository. I assumed it would come into main repository.

Should vagga install php7 from the community repository or keep php5 from main?

livioribeiro:
This pull request fixes php installation when running alpine 3.5 where php7 is available but the default is still php5. Previous implementation assumed php7 would be the default for alpine 3.5.

This also improves the usage of different runtimes by creating a symlink for the provided runtime at the default runtime location (`/usr/bin/php`) if it not exists. This is needed because php packages expect the binary `php` to be on PATH. While some OS packages already do this (e.g hhvm on Ubuntu Xenial) some do not (e.g php7 in alpine 3.5)

closes #385

### DIFF
--- a/docs/build_commands.rst
+++ b/docs/build_commands.rst
@@ -507,7 +507,14 @@ runtime:
     - !Ubuntu xenial
     - !ComposerConfig
       install_runtime: false
-      runtime_exe: hhvm
+      runtime_exe: /usr/bin/hhvm
+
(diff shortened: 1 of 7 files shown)


tailhook:
Looks good. Thank you!

PR #396, opened 2017-02-10
WIP: First prototype of `mixins` functionality

tailhook:
This is a bit smarter way to include another vagga config:

1. Comparing to `!SubConfig` includes full containers and commands
2. Comparing to `!*Include` allows to keep `contianers` and `commands` in the same file
3. Comparing to some super-smart version of `!*Include` can skip erroneous configs (useful if you generate parts of config)
4. Comparing to `!SubConfig` that generates `vagga.yaml` requires included configs to be generated in advance.

The biggest disadvantage of this pull request is obviously a feature creep. But also that containers and commands may come from different configs which and may clash and be overridden in non-obvious ways (typical problem with inheritance).

/cc @zemlanin, @popravich, @anti-social

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -20,6 +20,7 @@ use super::version::MinimumVagga;
 #[derive(RustcDecodable)]
 pub struct Config {
     pub minimum_vagga: Option<String>,
+    pub mixins: Vec<String>,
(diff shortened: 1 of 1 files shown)


tailhook:
Pushed into master.

PR #398, opened 2017-03-12
Zsh completion

uvNikita:
Features:

  - can complete standard vagga options (`--environ`, `--version` etc)
  - can complete both project and standard vagga commands
  - hidden commands (the ones that start with an underscore) are not showing by default, but will complete after writing first underscore: `_<TAB>`
  - can complete options from the commands with `options` (`docopt`) field specified in `vagga.yaml`
  - show description for both commands and options if available
  - can complete special objects (`FILE`, `DIR` and `PATH`) when specified (e.g. `--option=PATH`). This is provided by zsh `_arguments` function.


completions/zsh-completion.sh
```diff
@@ -0,0 +1,27 @@
...
+
+    # Show hidden options only when underscore is typed
+    if [[ $words[2] = '_'* ]]; then
+        listopts="--all"
+    fi
+
+    # Check if in folder with correct vagga.yaml file
+    vagga _list 1>/dev/null 2>/dev/null
+    if [ $? -eq 0 ]; then
+        IFS=$'\n' cmds=($(vagga _list $listopts | sed -e "/^ /d" | awk '{cmds=$1; $1=""; print cmds ":" $0}'))
```
tailhook: maybe `vagga _list --zsh` ?

tailhook:
Overall, looks too simple to be true :)

You should fix `install.sh` script. (moved bash completion, add zsh completion).

And I don't see the part that completes options. Is it somehow provided by zsh itself?

uvNikita:
- Refactored `print_list` function and added `--zsh` option to `_list` command
- Fixed `install.sh` script for bash completion. It seems zsh uses the `fpath` environ variable to find completion functions, not sure where to put `zsh-completion.sh` during installation.

> And I don't see the part that completes options. Is it somehow provided by zsh itself?

Yes, the last argument to `_arguments` function (`--`) tells zsh that it should run the command with `--help` and parse an output. We need to call this function twice: ones for vagga options itself (`vagga --help`), and once for command (`vagga _help <CMD> --help`).

tailhook:
I guess it should go into `/usr/share/zsh/site-functions`

uvNikita:
Put completion script to `/usr/share/zsh/site-functions` in installation script, however according to this [thread](http://www.zsh.org/mla/workers/2013/msg00622.html) different Linux distributions have different default folders for completions (Ubuntu/Debian -- `/usr/**local**/share/zsh/site-functions`, CentOS/RHEL -- `/usr/share/zsh/site-functions`).

I can also make PR in [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) with this script in case someone wants to install it in this way.

tailhook:
Looks good now. Thanks!

We need to wait for a release of vagga 0.7.1 before submitting it to oh-my-zsh I think.

PR #400, opened 2017-03-19
Improve zsh completion

uvNikita:
- Complete options of built-in commands in zsh
- Complete containers for _run command in zsh

Unfortunately, couldn't figure out how to complete options for `_run` command for now :(


tailhook:
Looks good. Thanks!

PR #401, opened 2017-03-21
Refactor PHP/Laravel tutorial

livioribeiro:
This is a refatctor of the Laravel tutorial, the changes are:

- update Laravel version
- change to alpine v3.5
- add better explanations about what is being done in the main container
- use mounts to avoid hacking the application to find the composer dependencies
- use redis instead of memcached
- improve adminer container
- remove section about deploying to shared server since Laravel is not designed for this kind of deployment

The change from ubuntu to alpine is considerable and I can rollback if needed


tailhook:
Great work. Thanks!

PR #402, opened 2017-03-27
WIP: yarn support

tailhook:
* [x] docs
* [ ] when `node_modules` are present in directory, yarn uses it instead of `--modules-folder` (warn?/create a signature file?)
* [ ] rename `YarnDependencies` -> `Yarn` (?)
* [ ] `YarnInstall` which does `yarn global add` (?)
* [x] tests
* [x] `flat: true` (?)
* [x] need to build container twice when there are no lockfile (was old vagga bug)
* [x] we don't depend on `package.json` but depend on `yarn.lock` which isn't too good (since we already have an update mechanism for lock, we are probably going to hash both files to be more strict with removing dependency back)
* [ ] how to do `yarn add`? Docs are enough?
* [x] install optional dependencies, add the option

### DIFF
--- a/docs/build_steps.rst
+++ b/docs/build_steps.rst
@@ -1122,6 +1122,34 @@ Node.JS Commands
         Setting the option to ``false`` is useful for setting up custom
         version of the node.js.
 
+.. step:: YarnDependencies
+
+   Works similarly to :step:`NpmDependencies` but installs packages using yarn
+   aand For example::
+
+        - !YarnDependencies
+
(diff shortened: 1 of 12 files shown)


tailhook:
Hm, the interesting fact is that when we remove something from `package.json` it doesn't get removed from `yarn.lock`. But if we're trying to install `package.json`, the package isn't installed.

For now, I think this is harmless enough for our use case.

tailhook:
Okay, the first version is landed. Moved remaining issues to #415

PR #414, opened 2017-04-14
Implement --pre option for pip in PipConfig

Orhideous:
This allow install pre-release and development versions of python packages


tailhook:
Looks fine. Thanks!

Issue #389, opened 2017-01-19
More powerful Copy and Depends


tailhook:
We want something like this:
```
!Copy
source: /work
rules:
- "*.py"
- "public/js/**/*.js"
- "/config"
- "!*~"
- "!*.old"
path: /some/dest
```

I.e. a list of rules similar to `.gitignore` rules, but reverse

/cc @anti-social

Orhideous:
Will this feature imprement recursive `!Depends` on entire folder?

anti-social:
We decided to support only absolute paths for including rules. It'll allow us to ignore non-matching directories.

PR #417, opened 2017-05-03
Support glob rules for Copy and Depends commands #389

anti-social:
I would like to validate exclusive options when parsing `vagga.yaml` but I don't know how to do that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ todo.txt
 /alpine-keys.apk
 /alpine
 
-/target
+target
 
 /tests/image/images
 /tests/image/nginx
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ script:
     for test in ${tests[@]}; do
       if [[ $test = cargo ]]; then
         vagga -E VAGGA_VERSION=v0.0.0-test _run rust-musl cargo test
+        (cd path_filter; vagga _run rust-musl cargo test)
       else
         vagga \
           -E UBUNTU_MIRROR=http://mirrors.us.kernel.org/ubuntu/ \
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,7 @@ script:
     tests=($(echo $TESTS))
     for test in ${tests[@]}; do
       if [[ $test = cargo ]]; then
-        vagga -E VAGGA_VERSION=v0.0.0-test _run rust-musl cargo test
-        (cd path_filter; vagga _run rust-musl cargo test)
+        vagga -E VAGGA_VERSION=v0.0.0-test _run rust-musl cargo test --all
       else
         vagga \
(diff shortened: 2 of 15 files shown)


.travis.yml
```diff
@@ -41,6 +41,7 @@ script:
     for test in ${tests[@]}; do
       if [[ $test = cargo ]]; then
         vagga -E VAGGA_VERSION=v0.0.0-test _run rust-musl cargo test
```
tailhook: I think there is "workspaces" stuff for that and `cargo test --all`

docs/build_steps.rst
```diff
@@ -682,6 +682,40 @@ Files and Directories
      unmapped users (the ones that don't belong to user's subuid/subgid range),
      will be set to ``nobody`` (65535).
 
+   rules
+     Leverages glob patterns instead of regular expressions to match paths.
+     This option conflicts with ``ignore-regex`` and ``include-regex`` options.
+
+     The rules are similar to those which is used in
```
tailhook: They are similar but the meaning of include/exclude rules is opposite.

docs/build_steps.rst
```diff
@@ -682,6 +682,40 @@ Files and Directories
...
+       directory even if it is empty.
+
+     By default there are some ignore rules that correspond ``ignore-regex``
+     expression:
+       ``["!.git/", "!.hg/", "!.svn/", "!.vagga/", "!*.bak", "!*.orig", "!*~",
+       "!#*#", "!.#*"]``
+
+   no-default-rules
+     Disables default rules. The option only works in pair with ``rules``
+     option.
```
tailhook: It's unclear what "default rules" you're talking about.

src/builder/error.rs
```diff
@@ -68,11 +68,11 @@ quick_error! {
             description("can't read directory")
             display("error reading directory: {:?}", errors)
         }
-        /// Can't compile regex
-        Regex(e: regex::Error) {
+        /// Can't read directory for copying
+        PathFilter(errors: Vec<path_filter::FilterError>) {
```
tailhook: What is the list of errors? Why don't we stop on the first error? Is there any cap on the number of errors here?

If there might be an error per file or something, it may consume lots of memory. While I don't think we're too memory-bound, but it's  much better to bail out on first error than to crash on OOM or hang by swapping when lots of errors being accumulated but not printed. Also note, that by default vagga-box has quite small amount of memory.

.gitignore
```diff
@@ -17,7 +17,7 @@ todo.txt
 /alpine-keys.apk
 /alpine
 
-/target
```
tailhook: Let's revert this one. It's not needed with workspaces and I believe it may be a problem because of ignoring such a generic name.

tailhook:
Overall, looks good enough. Let's merge it in and see how well it works. Thanks!

PR #428, opened 2017-06-29
Reverted capsule alpine to v3.5 due to bug in wget

anti-social:
Wget 1.26.2 closes write end of the socket, some http servers after that closes connection: [Bug 9471](https://bugs.busybox.net/show_bug.cgi?id=9471)


tailhook:
Merged in. Thanks!

PR #433, opened 2017-07-03
Implement _hardlink & _verify commands, global hardlinking

anti-social:


tailhook:
Have run `_hardlink --global` on my system:
```
WARN:vagga::wrapper::hardlink: Found and linked 1462331 (43GB) identical files
```
.. in 27 minutes. Which freed `18Gb` disk space. Which is pretty awesome.

The question is: can it report amount of space freed rather than amount of files scanned? (or both)

----

I'll play around with it a little bit before merging.

anti-social:
> in 27 minutes
I think you used debug build. I tested release build and it was quite fast.

tailhook:
It's release build. 43Gb, 1.5kk files looks reasonable I think. Also, some images (I don't know how many) were indexed in the process.

anti-social:
> The question is: can it report amount of space freed rather than amount of files scanned? (or both)

This is because of you already have been using hardlinking. So for example if you have 3 files `ino1, ino2, ino2` we create 2 hardlinks and count size twice. I will try to fix that.

anti-social:
> It's release build. 43Gb, 1.5kk files looks reasonable I think. Also, some images (I don't know how many) were indexed in the process.

Yeah, I tested that when we didn't check real file hash before hardlinking.

tailhook:
204 seconds second time (when no files are linked), but there were errors

anti-social:
Updated.

PR #435, opened 2017-07-06
Fixed config location for some internal commands

anti-social:
Possibly we should print warning when `.vagga/vagga.yaml` is used. Or print info message what config file was found.

### DIFF
--- a/src/builder/mod.rs
+++ b/src/builder/mod.rs
@@ -1,9 +1,9 @@
-use std::io::{stdout, stderr};
+use std::io::{stdout, stderr, Write};
 use std::default::Default;
 use std::path::Path;
 use rand;
 
-use config::read_config;
+use config::find_config;
 use config::{Config, Container, Settings};
 use argparse::{ArgumentParser, Store, StoreTrue};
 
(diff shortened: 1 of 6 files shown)


tailhook:
I think we should print INFO-level message. `.vagga/vagga.yaml` is specifically meant to locally override `vagga.yaml`.

It looks like it never worked since vagga 0.2, but still may be useful for testing new features or adding a git-ignored vagga.yaml.

src/builder/mod.rs
```diff
@@ -71,11 +71,16 @@ pub fn run(input_args: Vec<String>) -> i32 {
         }
     }
 
-    // TODO(tailhook) read also config from /work/.vagga/vagga.yaml
-    let config = read_config(&Path::new("/work/vagga.yaml")).ok()
-        .expect("Error parsing configuration file");  // TODO
+    let config = match find_config(&Path::new("/work"), false) {
```
tailhook: Please make a function that exits. It's kinda crash anyway.

src/builder/mod.rs
```diff
@@ -71,11 +71,16 @@ pub fn run(input_args: Vec<String>) -> i32 {
...
+        Ok((cfg, _)) => cfg,
+        Err(e) => {
+            writeln!(&mut stderr(), "Error parsing configuration file: {}", e)
```
tailhook: This message should include an explanation:
```
Internal error: Error parsing configuration file: {}. It usually happens either if configuration file was changed after vagga have been started or if permissions of the config file or some of the included files are wrong.
```

tailhook:
Looks good now. Thanks!

PR #440, opened 2017-07-13
Mount /etc/resolv.conf and /etc/hosts instead of copying

anti-social:


tailhook:
Notes to myself: technically it's a breaking change. There were folks who somehow abused two facts:

1. Container has some reasolv.conf which might still work
2. When you run second container, already running container got new resolv.conf

Both hacks are quite unreliable. And second one might also partially work: if resolv.conf was "overwritten" rather than "replaced". In the latter case in may also work as a bugfix, still not 100% reliable though. Another note is that in vagga box we overwrite resolv.conf of the virtualbox on each start of a new container.

Issue #441, opened 2017-07-18
GitDescribe command


anti-social:
It should be useful when building deploy containers.

Full version of the command:

```yaml
- !GitDescribe
  repo: /work
  path: /app_version
```

Also there can be short version:

```yaml
- !GitDescribe /app_version
```

Whether we need `git describe` options? If so what options should be implemented?

tailhook:
I'm not sure about `path:`. While we use it as destination path in copy-like commands, it looks super-confusing here. `destination-file` or just `file` is ok. Maybe we can also have empty `!GitDescribe` that depends on version but doesn't write the version anywhere.

PR #442, opened 2017-07-18
GitDescribe command #441

anti-social:


tailhook:
Looks great. Except we put `kebab-case` not `snake_case` in docs (both work though)

PR #445, opened 2017-07-28
Copy command now set files' modification time to the first second from the unix epoch

anti-social:
Should we set modification time for subcontainer commands and container snapshots?


tailhook:
Looks good. Thanks! And please add `[breaking]` to the title for such commits.

PR #447, opened 2017-08-05
Improve container build flow if cache url exists

### LINKED ISSUE #418: Handle existing image cache with interihance nicer
Given one container dependent on another (say `test` inherits `app`) vagga goes for building 'em recursively.
And somewhere in the deep of that build process it notices that there's an option to use image from cache (using `image-cache-url`).
So it goes for unpacking that image.

The problem is that it ends up trying to unpack (best case) or build (worst case) each of dependent containers.

It might just take the target containers cached archive and unpack it only instead of going down the chain.
This would save for 50% time building target container for 2 dependent containers and even more time in case when there's more.

anxolerd:
https://github.com/tailhook/vagga/issues/418

If cache_image_url present try to download cached image first instead of
building dependencies. If downloading cached image failed then return to
normal flow with building dependencies and building image itself
afterwards.

Attempt to download cache first occurs only if neither
--no-image-download nor --force flags were passed to build command.
Moreover, if the first attempt to download image from image_cache_url
failed, another one will not happen during the same build.

Closes #418

### DIFF
--- a/src/wrapper/build.rs
+++ b/src/wrapper/build.rs
@@ -709,6 +709,22 @@ pub fn build_wrapper(name: &str, force: bool, no_image: bool, wrapper: &Wrapper)
 {
     let container = wrapper.config.containers.get(name)
         .ok_or(format!("Container {:?} not found", name))?;
+
+    let mut skip_image_download = no_image;
+    if !no_image && !force && container.image_cache_url.is_some() {
+        debug!("Try to build only container {}", name);
+        let cont_info = ContainerInfo::new(name, container, force, no_image);
+        let build_res = build_container(&cont_info, wrapper);
+        if build_res.is_ok() {
+            return build_res;
+        }
+        unmount_service_dirs()?;
+        // we failed to download image, do not try again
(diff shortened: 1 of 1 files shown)


tailhook:
Well, the original problem isn't that easy:
1. Building container twice on any error isn't a good idea. There are tons of reasons why building may fail. And in most of them you don't want to repeat.
2. In some cases container's version depends on building the dependent container, so we may skip downloading dependency only when versioning code did not fail.

So the real patch would be:
1. Do `get_version_hash` if it suceeds, try to download cache, if that fails run whole build chain.
2. The `get_version_hash` call must be deduplicated (it's already called twice, calling it third time is no go)
3. And this should be done keeping proper locking of build dirs (this patch cheats by just calling `build_container` twice, but it's very annoying to needlessly wait for two attempts, and it also calls `get_version_hash` 4 times, which is very slow for some containers).

anxolerd:
I tried several times but failed to comply with requirements. Moreover, this PR is quite old and has merge conflicts. Thus I'd close it.

PR #450, opened 2017-08-15
Add clickhouse example

alxpy:


tailhook:
Thanks! It's also a good idea to add examples to docs: https://github.com/tailhook/vagga/tree/master/docs/examples/db

PR #460, opened 2017-10-04
vagga --version should now work without config

anti-social:


tailhook:
I've pushed a different fix. The idea is that with your patch `vagga --ignore-owner-check --version` (or any other option in between) doesn't work. While this specific combination of options doesn't look very useful, it may happen to be called if `alias vagga="vagga --ignore-owner-check"` is set or in similar situations.

PR #462, opened 2017-10-20
Added kafka example

anti-social:


tailhook:
Its also good idea to add a page to the docs

Issue #468, opened 2017-11-03
could not open USB device (insufficient permissions)


rainboh:
I created a simple container with installed sane-utils package

~~~
containers:
  base:
    setup:
      - !Ubuntu yakkety
      - !Install [sane-utils]

commands:
  bash: !Command
    container: base
    run: |
      bash
~~~

Then inside container I run command "sane-find-scanner" which outputs

~~~
  # sane-find-scanner will now attempt to detect your scanner. If the
  # result is different from what you expected, first make sure your
  # scanner is powered up and properly connected to your computer.

  # No SCSI scanners found. If you expected something different, make sure that
  # you have loaded a kernel SCSI driver for your SCSI adapter.

could not open USB device 0x05e3/0x0716 at 002:002: Access denied (insufficient permissions)
could not open USB device 0x1d6b/0x0002 at 002:001: Access denied (insufficient permissions)
could not open USB device 0x1d6b/0x0001 at 008:001: Access denied (insufficient permissions)
could not open USB device 0x1d6b/0x0002 at 001:001: Access denied (insufficient permissions)
...
~~~

checking permissions with "ls -la /dev/bus/usb/001/" shows
~~~
total 0
drwxr-xr-x  2 nobody nogroup     60 Nov  3 09:24 ./
drwxr-xr-x 12 nobody nogroup    240 Nov  3 09:24 ../
crw-rw-r--  1 nobody nogroup 189, 0 Nov  3 09:24 001
~~~

How can I fix the "insufficient permissions" ?

Thanks for your advice

tailhook:
I think you can do `sudo chown your_user /dev/bus/usb/001/*` (on a host system). If that works you can make an udev rule so that you don't need to do chown on plug-in/plug-out of a scanner (picking a random link from google: [this link](http://robots.mobilerobots.com/wiki/Linux_udev_USB_Device_Permissions_Configuration))

rainboh:
Creating an udev rule on the **host** which sets MODE="0777" for my scanner resolved the permission issue. 

First I tried the wrong way by adding the udev rule inside the container.

Maybe adding a hint to chapter "Tips and Tricks" in the docs might help other users too.

Still I have another issue as "simple-scan" is crashing with a segfault. But original issue is resolved for me.

Thanks for your quick help.

tailhook:
I'm glad it worked!

> Maybe adding a hint to chapter "Tips and Tricks" in the docs might help other users too.

[...]

PR #474, opened 2017-11-17
added hint howto fix usb device permissions to docs

rainboh:
As discussed in https://github.com/tailhook/vagga/issues/468#issuecomment-341890344 I tried to phrase a hint howto fix usb device permissions


tailhook:
Looks great. Thank you!

PR #476, opened 2017-12-14
Updated installing on Arch Linux

anti-social:


tailhook:
Thanks!

tailhook:
@anti-social, could you also add a small note at the bottom of how to switch the kernel back, like:
1. Update the kernel
2. Reboot and choose stock kernel
3. Remove `[linux-user-ns]` repo from pacman.conf
4. Remove `linux-user-ns` package
5. Run grub mkconfig (so absent kernel removed)

I'm not sure about exact sequence and command lines, but users will need it to make the switch easy.

PR #480, opened 2017-12-27
busybox apk is outdated

rainboh:
fetch is failing as 'busybox-static-1.25.1-r0.apk' is no longer available under http://dl-cdn.alpinelinux.org/alpine/v3.5/main/x86_64/


tailhook:
Thanks!

We should eventually upgrade to alpine v3.7 though.

rainboh:
CI was failing ?

tailhook:
That's OSX version, it's kinda tier2 platform for now, because it's dysfunctional yet (but was compiling some time ago). I'm keeping it to have more motivation to work on it from time to time. I might need to add `allow_failure` to make it less surprising though :)

PR #481, opened 2017-12-27
support fetching binaries for different target arch

rainboh:
Extended the script to allow fetching binaries for other targets (in my case armhf).

By default 'x86_64' binaries are fetched if no parameter is provided to the script.
To fetch e.g. armhf' binaries
~~~
fetch_binaries.sh armhf
~~~

If you prefer I can change the script to force user for always providing 'arch' parameter, and print usage hint in case he forgot to do so.

Still on alpine v3.5.
You want me to upgrade to v3.7 ?

### DIFF
--- a/fetch_binaries.sh
+++ b/fetch_binaries.sh
@@ -5,26 +5,37 @@ APK_TOOLS=apk-tools-static-2.6.9-r0.apk
 BUSYBOX=busybox-static-1.25.1-r1.apk
 ALPINE_KEYS=alpine-keys-1.3-r0.apk
 
+ARCH=${1:-x86_64}
 
-mkdir alpine 2>/dev/null || true
-cd alpine
+SHA1SUMS_x86_64="\
+31b29926d6be7efb389b49d9d53b557e9b25eb7c  $APK_TOOLS
+8b5639a22af1e656f03931ab38ef51285b3b9dd2  $BUSYBOX
+f1c6e5f7209885fec5c8dd8c99446036852988a0  $ALPINE_KEYS"
+
+SHA1SUMS_armhf="\
+5d21cc3b2641bdb47231c323e9f5d736324ed425  $APK_TOOLS
+63201684a82918bf3d92c122c0679681ec880bf6  $BUSYBOX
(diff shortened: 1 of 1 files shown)


tailhook:
> If you prefer I can change the script to force user for always providing 'arch' parameter, and print usage hint in case he forgot to do so.

Current solution is fine. We expect most users install vagga from package, not compile. So we can change it later if it's not convenient anymore (for now I expect most users use x86_64)

> You want me to upgrade to v3.7 ?

You can make a (separate) PR, if you wish. Otherwise, I'll do it eventually :)

tailhook:
Merged. Thanks!

What is a status overall, does vagga at least start?

rainboh:
Vagga is such a nice and great piece of software and already well prepared for different targets.

It's not only starting.
For "target = arm-unknown-linux-musleabihf" it works as expected and is useable on my pi2.

~~~
~/vagga/test $ /usr/lib/vagga/vagga bash
root@black-pearl:~# uname -a
Linux black-pearl 4.9.52-v7+ #1038 SMP Fri Sep 29 16:26:52 BST 2017 armv7l armv7l armv7l GNU/Linux
root@black-pearl:~# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.3 LTS"
root@black-pearl:~#
~~~

To get it compiled there is just one more minor change needed in vagga src file "file_util.rs". You will get a PR soon :-)

Ok. When x-compiling there are still building issues which I don't know how to resolve yet:
- libgit2 produces a link error as cmake thinks qsort_r is available for the arm target
- release build for "target = arm-unknown-linux-musleabihf" is failing
- build for "target = armv7-unknown-linux-musleabihf" is failing

Guess my knowledge about x-compile is still to minor to get all this stuff resolved.
But nevertheless I'm happy the "arm-unknown-linux-musleabihf debug build" of vagga runs on my pi.
Thank you for vagga.

PR #482, opened 2017-12-27
support target arm

rainboh:
changed types to be compatibe with target "arm-unknown-linux-musleabihf"

### DIFF
--- a/src/file_util.rs
+++ b/src/file_util.rs
@@ -10,13 +10,13 @@ use std::os::unix::io::AsRawFd;
 use std::os::unix::ffi::OsStrExt;
 
 use nix;
-use libc::{uid_t, gid_t, c_int, utime, utimbuf};
+use libc::{uid_t, gid_t, c_int, utime, utimbuf, c_char, time_t};
 use nix::fcntl::{flock, FlockArg};
 
 use path_util::ToCString;
 
 extern "C" {
-    fn lchown(path: *const i8, owner: uid_t, group: gid_t) -> c_int;
+    fn lchown(path: *const c_char, owner: uid_t, group: gid_t) -> c_int;
 }
 
 pub struct Lock {
@@ -471,8 +471,8 @@ pub fn set_times<P: AsRef<Path>>(path: P, atime: i64, mtime: i64)
 {
     let filename = CString::new(path.as_ref().as_os_str().as_bytes()).unwrap();
     let utimes = utimbuf {
-        actime: atime,
-        modtime: mtime,
+        actime: atime as time_t,
+        modtime: mtime as time_t,
     };
     let rc = unsafe { utime(filename.as_ptr(), &utimes) };
     if rc != 0 {


tailhook:
Well, I think we can use [`lchown`](https://docs.rs/libc/0.2.34/libc/fn.lchown.html) from libc? I think the reason we have an `extern` declaration` is that function was absent in some old version of libc.

rainboh:
Good suggestion. Changed it to use llibc::lchown

rainboh:
Btw. As mentioned in   https://github.com/tailhook/vagga/pull/481#issuecomment-354136190  this is the only change needed in vagga repo so far (), to get debug build with "target=arm-unknown-linux-musleabihf" working on pi2

tailhook:
Merged. Thanks!

> Btw. As mentioned in #481 (comment) this is the only change needed in vagga repo so far ()

You mean that it works if compiling on PI itself. All other problems mentioned there is because of cross-compiling, right?

rainboh:
> You mean that it works if compiling on Pi itself.

No. On Pi itself all dependent crates get compiled but when cargo comes to 'vagga' the system gets overloaded. I tried but then I reseted the Pi after some hours because I was to lazy to wait. Guess linking all crates is too much for Pi.

> All other problems mentioned there is because of cross-compiling, right?

Correct. But so far I can only workaround the qsort_r issue by patching the libgit2 code.

Only working setup at moment is:
- patch libgit 2
- x-compile for "target=armv7-unknown-linux-musleabihf" (no release build)

But the created vagga executable works fine. No strange issues detected so far when playing around.

Of course 'container building' speed on Pi is slow. Don't know if it is as to be expected, or if it can be tuned somehow.

tailhook:
Aha, okay, I don't think there are many people who build on Pi, so I don't think it matters.

Can you make a pull request to libgit2? We can feature-gate libgit2 if needed but probably PR is the best.

rainboh:
> Can you make a pull request to libgit2?

No. I commented a check which is no proper solution. If I make a PR I don't expect it will get merged.


If I compile vagga with
~~~
cargo build --target=arm-unknown-linux-musleabihf
~~~
crate libz-sys (needed by libgit2) fails.


If I compile vagga with
~~~
CC=arm-linux-gnueabihf-gcc cargo build --target=arm-unknown-linux-musleabihf
~~~
linking fails because 'qsort_r' is missing (which seems to be expected for musl https://github.com/libgit2/libgit2/issues/3438#issuecomment-143575537)


First I need to figure out if I do the wright thing, means 'Is it correct if I set CC for cargo build ?'

I opened issue https://github.com/alexcrichton/libz-sys/issues/35
Let's hope that  I asked the right questions.

tailhook:
> First I need to figure out if I do the wright thing, means 'Is it correct if I set CC for cargo build ?'

I guess yes. While libz-sys cargo could detect it, linker will fail in the same way when libz will successfully compile.

> linking fails because 'qsort_r' is missing (which seems to be expected for musl libgit2/libgit2#3438 (comment))

Looks like it's because you use gnu headers instead of musl ones by default. Here is how I can compile it: https://github.com/tailhook/vagga/pull/483 (nowhere to test though)

rainboh:
> Looks like it's because you use gnu headers instead of musl ones by default.

You are right. Source code is ready for 'arm'. Looks like issue is setting up correct header/lib path for compile/linking.

Meanwhile I freed up some space on SD and mem on my Pi and retried to build on Pi itself where I expect correct settings for headers and libs. I tried :

~~~
1. cargo build --target=arm-unknown-linux-musleabihf
2. CC=musl-gcc cargo build --target=arm-unknown-linux-musleabihf
3. CC=musl-gcc cargo build --target=arm-unknown-linux-musleabihf --release
~~~

1. fails with same 'qsort_r' linking error
2. success
3. fails because linker run out of mem. I will later spend some swap space and retry.

I expected 1. to be successfull too because on my x86_64 it's sufficient to
~~~
cargo build --target=x86_64-unknown-linux-musl
~~~
while on Pi I need to add 'CC=musl-gcc'.

> Here is how I can compile it: #483 (nowhere to test though)

Fails for me. I left comment on other thread

tailhook:
> fails with same 'qsort_r' linking error

Different headers. You can also use `CFLAGS=-I/usr/include/arm-linux-musleabihf` but `CC` is also okay probably

> fails because linker run out of mem. I will later spend some swap space and retry.

You can also try disable `LTO`, make less number of threads, and `opt-level = "s"` instead of default 3. Or maybe just lower it.

Do you think providing a package for arm is useful?

rainboh:
> You can also use CFLAGS=-I/usr/include/arm-linux-musleabihf but CC is also okay probably

Correct. Setting include path instead of CC is also building properly on Pi

~~~
CFLAGS=-I/usr/include/arm-linux-musleabihf cargo build --target=arm-unknown-linux-musleabihf
~~~

> Do you think providing a package for arm is useful?

Of course a package would make life easier.
Sorry, I have no knowledge about package building and can't support here.

No idea how many people have a use case for vagga on their Pi (or other arm devices). On the other side if people know it's possible they will start to think about and usage will increase.

As it is a matter of priority and time I would say arm package can wait (unless it's almost no effort for you to implement needed build steps). CI is more important. And maybe some hints 'howto use vagga on arm'.

Kochise:
Fyi, `qsort_r` has been added to Alpine on 2021/09/22 (musl-1.2.2-r7) :

https://gitlab.alpinelinux.org/alpine/aports/-/commit/bf5bbfdbf780092f387b7abe401fbfceda90c84d

Profit.

Issue #478, opened 2017-12-20
Add `signed-by` to UbuntuRepo


tailhook:
New debian-based distributions support the following:
```
deb [signed-by=6809F110790E0720856B562F744ACF4DF3319FFA] https://deriv.example.net/debian/ stable main
```
This is both, more secure, and allows to skip `apt-key adv --recv-key` (the latter also requires additional package on newer distros).

This is a bit more complex for UbuntuPPA though. Should we deprecate UbuntuPPA, or add microformat for key too?

PR #485, opened 2018-01-06
use port 80 to connect to ubuntu keyserver

rainboh:
At office I got an error when I use !AptTrust

According to this info https://gpgtools.tenderapp.com/kb/faq/cant-reach-key-server-are-you-behind-a-company-firewall it's because port 11371 is blocked.

~~~
# this was failing
!AptTrust keys: [93C4A3FD7BB9C367]
# this did working
!Sh
apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 93C4A3FD7BB9C367
~~~

As port 80 should usually not be blocked, I changed default keyserver port in this PR.
Did not yet test this change in office, but expect it to work.

Let me know if you prefer port 443

### DIFF
--- a/src/builder/commands/ubuntu.rs
+++ b/src/builder/commands/ubuntu.rs
@@ -110,6 +110,12 @@ impl AptTrust {
 }
 
 #[derive(Debug, PartialEq)]
+enum AptHpks {
+    No,
+    Installed,
+}
+
+#[derive(Debug, PartialEq)]
 enum AptHttps {
     No,
     Need,
@@ -129,6 +135,7 @@ pub struct Distro {
     codename: Option<String>,
     apt_update: bool,
     apt_https: AptHttps,
+    apt_hpks: AptHpks,
     has_indices: bool,
     has_universe: bool,
     eatmydata: EatMyData,
@@ -189,15 +196,7 @@ impl Distribution for Distro {
     {
         if self.apt_update {
             if self.apt_https == AptHttps::Need {
-                let apt_https_deps = ["ca-certificates", "apt-transport-https"];
-                for dep in apt_https_deps.iter() {
-                    ctx.build_deps.insert(dep.to_string());
-                }
-                apt_get_update(ctx, &[
-                    "--no-list-cleanup",
-                    "-o", "Dir::Etc::sourcelist=sources.list",
-                    "-o", "Dir::Etc::sourceparts=-"])?;
-                apt_get_install(ctx, &apt_https_deps, &self.eatmydata)?;
+                self.install_apt_deps(ctx, &["ca-certificates", "apt-transport-https"])?;
                 self.apt_https = AptHttps::Installed;
             }
             if !self.has_indices {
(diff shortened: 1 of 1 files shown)


tailhook:
If 443 does work, it's better to use it.

Also, in the future we're going to implement #478 and deprecate AptTrust. It looks like it solves your problem too, right? Just I'm not sure a minimum ubuntu version which supports the option.

rainboh:
Reading doc https://wiki.debian.org/DebianRepository/UseThirdParty I understood that it always require to download a keyring file.

If  AptTrust gets deprecated this means to install repos which only provide the key but no keyring we still need to get the key by
~~~
!Sh
apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 93C4A3FD7BB9C367
~~~

Why not keep AptTrust in addition to support 'signed-by' ?

Do you know where to get keyring for ansible (https://launchpad.net/~ansible/+archive/ubuntu/ansible) ?

tailhook:
> Why not keep AptTrust in addition to support 'signed-by' ?

It will be kept anyway. Deprecation is not removal.

But it looks like I have understood what is the new way mistakenly in #478, and I'm not yet sure what the right way is. I've only seen few mentions [like this](https://bugs.launchpad.net/ubuntu/+source/enigmail/+bug/1623087) that it's deprecated.

Anyway, can you confirm that port 443 works for you? And we will merge this PR regarless of the future of `apt-key` :)

rainboh:
Port 80 works out of the box.

Port 443 ("hkps://keyserver.ubuntu.com:443") will only work when adding this to container
~~~
      - !UbuntuUniverse
      - !BuildDeps [gnupg-curl, ca-certificates]
~~~

If "gnupg-cur"l is not installed error "gpgkeys: HTTP fetch error 1: unsupported protocol" is reported.
Without "ca-certificates" error "gpgkeys: HTTP fetch error 77:" is reported.

This means switching to https will break existing containers unless vagga takes care to install this 2 packages.

Is that possible with something like ?
~~~
guard.ctx.build_deps.insert(packagename)
~~~
Your expertise is needed if https should be used.

tailhook:
We already have something similar [here](https://github.com/tailhook/vagga/blob/master/src/builder/commands/ubuntu.rs#L969-L976) and [here](https://github.com/tailhook/vagga/blob/master/src/builder/commands/ubuntu.rs#L191-L202) to use https-based repositories. These pieces might need to be factored out to a functions, but it shouldn't be too hard. Don't hesitate to ask for more help though.

rainboh:
Updated PR.

Adding same key twice is not detected, e.g. if this is inside vagga.yaml 
~~~
- !AptTrust keys: [93C4A3FD7BB9C367]
- !AptTrust keys: [93C4A3FD7BB9C367]
~~~
key will get downloaded twice, but required packages [gnupg-curl, ca-certificates] only downloaded and installed once.

tailhook:
Looks good. Thanks!

PR #493, opened 2018-05-17
Added pattern parameter to the GitDescribe build step

anti-social:


tailhook:
Merged!

(I think it's a bit unfortunate that we don't have `v*` by default, but anyway...)

Issue #465, opened 2017-11-01
BuildDeps: packages with version


anti-social:
```
containers:
  test-build-deps:
    setup:
    - !Ubuntu xenial
    - !BuildDeps [make=4.1-6]
```

Fails with:
```
...
E: Unable to locate package make=4.1-6                   
E: Couldn't find any package by glob 'make=4.1-6'                        
E: Couldn't find any package by regex 'make=4.1-6'
E: No packages found                                                  
ERROR:vagga::builder: Error building container "test-build-deps": error running <Command "/usr/bin/apt-mark" "auto" "make=4.1-6"; environ[5]; chroot="/vagga/root"; work-dir="/work"> exited with code 100
ERROR:vagga::wrapper: Error executing _build: Builder exited with code 1
Command <Command "/proc/self/exe" "__wrapper__" "_build" "test-build-deps"; environ[3]; uid_map=[UidMap { inside_uid: 0, outside_uid: 1000, count: 1 }, UidMap { inside_uid: 1, outside_uid: 100000, count: 65535 }]
; gid_map=[GidMap { inside_gid: 0, outside_gid: 100, count: 1 }, GidMap { inside_gid: 1, outside_gid: 100000, count: 65535 }]> exited with code 124
```

Should `BuildDeps` support specifying package version?

tailhook:
> Should BuildDeps support specifying package version?

Yes it should

tailhook:
Fixed in #465

PR #494, opened 2018-05-18
Fixed BuildDeps step with versioned packages #465

anti-social:


tailhook:
Looks good. Hopefully ubuntu test will not break soon...

PR #496, opened 2018-05-31
prevents truncation of long symbolic links when unpacking

rainboh:
Unpacking tar archive which contains symbolic links with long names produces broken links as long names get trancuted. This PR fixed issue for me.

Even worse the error get's not detected during unpacking. `symlink(&src, &path)` does not report error even if `src` does not exist.

Don't know if it will be ok to add check if `src` exist at time when symlink get created, because may be targeted `src` file is not yet unpacked. But would be nice to have some inegrity check (checksum ?) of unpacked archive as it leads to strange (and time consuming) errors.

### DIFF
--- a/src/builder/commands/tarcmd.rs
+++ b/src/builder/commands/tarcmd.rs
@@ -133,7 +133,7 @@ fn unpack_stream<F: Read>(file: F, srcpath: &Path, tgt: &Path,
             }
             dir_builder.create().map_err(&write_err)?;
         } else if entry.is_symlink() {
-            let src = src.header().link_name().map_err(&read_err)?
+            let src = src.link_name().map_err(&read_err)?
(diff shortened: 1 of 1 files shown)


tailhook:
Merged. Thanks!

> Don't know if it will be ok to add check if src exist at time when symlink get created, because may be targeted src file is not yet unpacked.

I'm not sure this is possible or desirable. As it's often fine if symlinks point to external files. And I don't think there is a way to check checksums over the actual unpacked files.

This is just a bug in our codebase, though, misusing the API. So I don't think we should have any deep changes because of it.

However, it may help to update a docstring of the `Header::link_name` in the `tar` crate to clearly indicate that it may return broken link names (or emit error?), because I remember I've fixed similar bug somewhere. (And maybe the unbroken API is newer than the usage in vagga, I'm not sure).

PR #506, opened 2018-11-23
update installation.txt

KarelWintersky:
with gentoo recipe

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -269,6 +269,25 @@ Also make sure you have musl-gcc in your path::
 Then just build using cargo and the appropriate target::
 
     $ cargo build --target x86_64-unknown-linux-musl
+    
+Build from source (Gentoo, docker)
+==========================
+
+Clone vagga repository
+
+   $ git clone https://github.com/tailhook/vagga && cd vagga
+   
+Describe vagga version
+
+   $ git describe 
+   
+For example: `v0.8.1-19-g372bded`
+
+Compile vagga with dockerized rust:
+
+   $ docker run --rm --user "$(id -u)":"$(id -g)" -e VAGGA_VERSION=v0.8.1-19-g372bded -v "$PWD":/usr/src/myapp -w /usr/src/myapp rust cargo build --release
+   
+Compiled binary move to `/usr/local/bin`
 
 
 OS X / Windows
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -270,8 +270,8 @@ Then just build using cargo and the appropriate target::
 
     $ cargo build --target x86_64-unknown-linux-musl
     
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -333,4 +352,4 @@ To run container with Ubuntu add ubuntu-miror to your vagga settings file
 In your vagga.yaml select proper architecture::
 
        setup:
-       - !UbuntuRelease { codename: xenial, arch: armhf }
```
tailhook: What is the change here? Whitespace?

docs/installation.rst
```diff
@@ -269,6 +269,25 @@ Also make sure you have musl-gcc in your path::
...
+   
+Describe vagga version
+
+   $ git describe 
+   
+For example: `v0.8.1-19-g372bded`
+
+Compile vagga with dockerized rust:
+
+   $ docker run --rm --user "$(id -u)":"$(id -g)" -e VAGGA_VERSION=v0.8.1-19-g372bded -v "$PWD":/usr/src/myapp -w /usr/src/myapp rust cargo build --release
```
tailhook: If you're already using `$(id -u)` you might want to use `-e VAGGA_VERSION=$(git describe)`, right?

docs/installation.rst
```diff
@@ -269,6 +269,25 @@ Also make sure you have musl-gcc in your path::
 Then just build using cargo and the appropriate target::
 
     $ cargo build --target x86_64-unknown-linux-musl
+    
+Build from source (Gentoo, docker)
```
tailhook: What's gentoo-specific here? Maybe just "Building From Source Using Docker"?

docs/installation.rst
```diff
@@ -269,6 +269,25 @@ Also make sure you have musl-gcc in your path::
...
+==========================
+
+Clone vagga repository
```
tailhook: I believe all these needs double-colon for proper formatting

docs/installation.rst
KarelWintersky: in my environment `-e VAGGA_VERSION=$(git describe)` does not work :(

docs/installation.rst
KarelWintersky: yeah, i think so.

docs/installation.rst
KarelWintersky: fixed

docs/installation.rst
KarelWintersky: I don't know, why it does not work, but without `-e VAGGA_VERSION=v0.8.1-19-g372bded` it drops with complilation error (due VAGGA_VERSION will be empty)

docs/installation.rst
KarelWintersky: fixed

PR #548, opened 2021-10-08
Initial github-actions-based tests

tailhook:


anti-social:
Think we should use vagga to build itself as for instance for #555 we'll need to compile `openssl` statically.

anti-social:
Have you any thoughts why does "image: pull & push" test fail? On my computer it passes.

tailhook:
> Think we should use vagga to build itself as for instance for #555 we'll need to compile `openssl` statically.

I'm not sure what is the difference.

> Have you any thoughts why does "image: pull & push" test fail? On my computer it passes.

No idea. We have to make more debugging in the test.

anti-social:
> I'm not sure what is the difference.

To not duplicate setup code from a build container into github-ci. For instance to install `openssl` we'll need to add something like:

```yaml
    # https://qiita.com/liubin/items/6c94f0b61f746c08b74c
    - !Sh |
        ln -s /usr/include/x86_64-linux-gnu/asm /usr/include/x86_64-linux-musl/asm
        ln -s /usr/include/asm-generic /usr/include/x86_64-linux-musl/asm-generic
        ln -s /usr/include/linux /usr/include/x86_64-linux-musl/linux
    - !TarInstall
      url: https://github.com/openssl/openssl/archive/OpenSSL_1_1_1l.tar.gz
      script: |
        CC="musl-gcc -fPIE -pie" \
          ./Configure no-shared no-async \
          --prefix=/usr/lib/x86_64-linux-musl \
          --openssldir=/usr/lib/x86_64-linux-musl/ssl \
          linux-x86_64
        make depend
        make -j$(nproc)
        make install_sw
```

But possibly it's not a problem at all.

tailhook:
> > I'm not sure what is the difference.
> 
> To not duplicate setup code from a build container into github-ci. For instance to install `openssl` we'll need to add something like:
> 

Yes, I think that's not a problem. Although, if rust-cache works for vagga-based builds, we can probably go either way.

PR #550, opened 2021-12-02
Fix building

anti-social:
Old digest libs depend on `crypto-mac` library which is missing on crates.io at the moment.


tailhook:
Nice. Thanks!

PR #553, opened 2021-12-05
Fix pip and its tests

anti-social:


src/process_util.rs
```diff
@@ -47,16 +47,33 @@ pub fn squash_stdio(cmd: &mut Command) -> Result<(), String> {
...
     cmd.stdout(Stdio::piped());
+    cmd.stderr(Stdio::piped());
     info!("Running {}", cmd_show(&cmd));
     let mut child = cmd.spawn()
         .map_err(|e| format!("{}", e))?;
-    let mut buf = Vec::with_capacity(1024);
-    child.stdout.take().unwrap().read_to_end(&mut buf)
-        .map_err(|e| format!("Error reading from pipe: {}", e))?;
+    let mut output = CapturedOutput::new();
+    child.stdout.take().unwrap().read_to_end(&mut output.stdout)
```
tailhook, line 71: The problem with this is that it's potentially blocking. I.e. if process have written more than (in-kernel) buffer size into stderr, and it needs to write to stderr before it will write and close stdout this will be a problem.

On the other hand, current Rust has `Command::output()` method that does the job. So this function is barely needed for better logging and more context for errors.

src/process_util.rs
```diff
@@ -47,16 +47,33 @@ pub fn squash_stdio(cmd: &mut Command) -> Result<(), String> {
...
+        Self {
+            stdout: Vec::with_capacity(1024),
+            stderr: Vec::with_capacity(1024),
+        }
+    }
+}
+
+pub fn capture_output(mut cmd: Command) -> Result<CapturedOutput, String> {
     cmd.stdout(Stdio::piped());
+    cmd.stderr(Stdio::piped());
```
tailhook, line 66: I'm not sure this is a good idea to use for every thing that was previously `capture_stdout`. The idea was that errors are visible. Here it will swallow the stderr output and only show error code.

tailhook (commented):
Except `capture_output` and `capture_stderr` should be different functions and `capture_output` using `Command::output` and used carefully, this looks good.

src/process_util.rs
anti-social: Oh, I see it now. It is really tricky case.
But here we use `unshare::Command`. Stdlib's `Command::output()` uses under the hood a private [read2](https://doc.rust-lang.org/src/std/sys/unix/pipe.rs.html#81) function.

Possible there should be an analogue functionality inside `unshare`. Or is it worth to make 2 functions for capturing `stdout` and `stderr`?

src/builder/commands/pip.rs
```diff
@@ -15,6 +18,10 @@ use builder::commands::generic::{run_command_at_env, capture_command};
 #[cfg(feature="containers")] use file_util::Dir;
 use build_step::{BuildStep, VersionError, StepError, Digest, Config, Guard};
 
+const PIP_HOME: &str = "/tmp/pip-install";
+const PYTHON_PATH: &str = concatcp!("/tmp/non-existent:", PIP_HOME);
```
anti-social, line 22: Why do we need `/tmp/non-existent` path here?

tailhook: This is some pip feature that tries to install something in the first directory in python path if that exists, or putting some `.pth` file there. I don't remember details unfortunately.

So if you try without it and rediscover the reason, please put the comment about that.

anti-social: Don't see anything strange with the current pip.

anti-social:
Ping @tailhook

src/process_util.rs
tailhook: Oh, I entirely forgot that we don't have that in unshare :(

Yes two functions is fine. Enum as you have done is fine too.

src/builder/commands/pip.rs
tailhook: Yes looks fine.

PR #554, opened 2021-12-08
Allow to pass pip dependencies to get-pip script

anti-social:
Some other variants to name it:
- `pip-dependencies`
- `get-pip-args`

### DIFF
--- a/tests/pip/vagga.yaml
+++ b/tests/pip/vagga.yaml
@@ -85,6 +85,15 @@ containers:
     - !Env { CFLAGS: "-I/usr/include/libxml2 -I/usr/include/libxslt" }
     - !Py3Install [lxml]
 
+  pip-deps:
+    setup:
+    - !Ubuntu focal
+    # Fails with following error if setuptools version is not specified:
+    # error in anyjson setup command: use_2to3 is invalid
+    - !PipConfig
+      pip-deps: [setuptools<58]
+    - !Py3Install [anyjson==0.3.3]
+
 commands:
   pty-output: !Command
     container: py3-ubuntu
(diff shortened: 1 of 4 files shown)


tailhook:
This is interesting: wasn't that `/non-existent` from #553, because this allowed pip not install setuptools (and you could set your own version in `requirements.txt`)? Is there something simple that requires old setuptools that I can test with?
 
Although, I'm not against this solution even if the above is true. It maybe somewhat more straightforward approach.

Let's name it `get-pip-args`. But keep the example with setuptools in docs, so it's more or less clean what it is for.

tests/pip/vagga.yaml
```diff
@@ -85,6 +85,15 @@ containers:
...
     - !Py3Install [lxml]
 
+  pip-deps:
+    setup:
+    - !Ubuntu focal
+    # Fails with following error if setuptools version is not specified:
+    # error in anyjson setup command: use_2to3 is invalid
+    - !PipConfig
+      pip-deps: [setuptools<58]
+    - !Py3Install [anyjson==0.3.3]
```
anti-social, line 95: It's an example of package that can't be installed with `setuptools>=59`. And it isn't trivial to fix the error. Following solution almost works:

```yaml
    - !Sh |
        export PYTHONPATH=/tmp/pip-install
        python3 -m pip uninstall -y setuptools
        python3 -m pip install "setuptools<58.0"
```

but requires to install at least one package before that.

anti-social: But you should check you don't have cached wheel.

anti-social:
There are a lot of issues: https://github.com/pypa/setuptools/issues?q=is%3Aissue+use_2to3

anti-social:
> because this allowed pip not install setuptools (and you could set your own version in requirements.txt)

At the moment `pip` always install its own `setuptools` into target path.

tailhook:
Okay. Let's just rename it to `get-pip-args` as we don't validate them anyway.

PR #567, opened 2021-12-18
Celan up explicit verbosity in tests

anti-social:
We do not need to print command's output explicitly as modern bats has some useful options: `--verbose-run`, `--show-output-of-passing-tests`, `--trace`


tailhook:
Wow, great!

PR #568, opened 2021-12-20
Apt cache overlay

anti-social:


src/builder/commands/ubuntu.rs
```diff
@@ -626,6 +645,46 @@ impl Distro {
...
+            for (entry, name) in iter {
+                if !name.ends_with(".deb") { continue }
+                if !entry.file_type()?.is_file() { continue }
+                let src = entry.path();
+                let tmp_prefix: String = thread_rng()
+                    .sample_iter(&Alphanumeric)
+                    .take(6)
+                    .map(char::from)
+                    .collect();
+                let tmp = cache_dir.join(format!(".{}-{}.part", tmp_prefix, name));
```
tailhook, line 671: This should make into a function. We already have enough ad-hoc ways to create a tmp filename.

src/builder/commands/ubuntu.rs
```diff
@@ -626,6 +645,46 @@ impl Distro {
...
+                let dst = cache_dir.join(name);
+                if !dst.exists() {
+                    // Cannot rename without copying due to "Cross-device link (os error 18)"
```
tailhook, line 674: Is this because it moves across mountpoints? Or because on your system they are different folders? I think we could do some tweak to make them in the same filesystem. Although, I'm not sure it's easy to do securely.

src/builder/context.rs
```diff
@@ -112,26 +117,102 @@ impl<'a> Context<'a> {
...
+    }
+
+    pub fn get_cached_dir(&self, path: &Path) -> Option<&CachedDir> {
+        self.cache_dirs.get(path)
+    }
+
+    pub fn is_cached_dir_overlay(&self, path: &Path) -> bool {
+        self.get_cached_dir(path)
+            .and_then(|d| d.overlay_dir.as_ref())
+            .is_none()
```
tailhook, line 215: I'm a bit confused. Should this be `is_some`?

src/builder/context.rs
```diff
@@ -112,26 +117,102 @@ impl<'a> Context<'a> {
...
     }
 
-    pub fn add_cache_dir(&mut self, path: &Path, name: String)
+    pub fn add_cache_dir(&mut self, path: &Path, name: &str)
         -> Result<(), String>
     {
+        let entry = match self.cache_dirs.entry(path.to_path_buf()) {
+            Entry::Vacant(entry) => entry,
+            Entry::Occupied(_) => return Ok(()),
+        };
```
tailhook, line 126: I think it overoptimizes it.
`if self.cache_dirs.contains_key(path) { return ...}`
.. with the insert at the end of the function should be clearer. Also it would avoid `to_path_buf` allocation on `return` path, so might be even better performance-wise.

src/builder/context.rs
```diff
@@ -112,26 +117,102 @@ impl<'a> Context<'a> {
...
+        Ok(())
+    }
+
+    pub fn add_cache_dir_overlay(&mut self, path: &Path, name: &str)
+        -> Result<(), String>
+    {
+        let entry = match self.cache_dirs.entry(path.to_path_buf()) {
+            Entry::Vacant(entry) => entry,
+            Entry::Occupied(_) => return Ok(()),
+        };
```
tailhook, line 161: Same think. `if !..contains_key` should be fine.

src/builder/context.rs
```diff
@@ -112,26 +117,102 @@ impl<'a> Context<'a> {
...
+
+        let overlay_dir = Path::new("/vagga/container").join(name);
+        let work_dir = overlay_dir.join("work");
+        let upper_dir = overlay_dir.join("upper");
```
tailhook, line 165: I don't think `upper` is a good name here. By "upper" in the previous discussion I meant upper from `/vagga/container/root` which happens to be `/vagga/container`.

Name should be something like `tmp_cache` or something.

Also we should probably have a subfolder, so that we can add more layered things, here:
```/vagga/container/tmp_cache/apt```

File: src/builder/commands/ubuntu.rs [hunk 7]
```diff
@@ -626,6 +645,46 @@ impl Distro {
...
+                    copy(&src, &tmp)?;
+                    copy_utime(&src, &tmp)?;
+                    rename(tmp, dst)?;
+                }
+            }
+            Ok(())
```
tailhook, line 680: Also we should cleanup the cache from the container i.e. `remove_dir_all(overlay_dir.join("tmp_cache"))` (that leaves only dirs we we can `rename` to the cache dir, but that's also nice to remove).

src/builder/context.rs
anti-social: Me too )))

src/builder/commands/ubuntu.rs
anti-social: Vagga cache and container is on the same filesystem. So I think it is because of different mount points. Will try to find out an exact reason.

src/builder/context.rs
anti-social: `upper` is just an overlayfs terminology. So anyone who have read overlayfs docs knows what it is for.

I like `/vagga/container/cache`.

src/builder/commands/ubuntu.rs
anti-social: Could you explain what do you want?

src/builder/commands/ubuntu.rs
anti-social: `man 2 rename`:
>     EXDEV  oldpath and newpath are not on the same mounted filesystem.  (Linux permits a filesystem 
>            to be mounted at multiple points, but rename() does not work across different mount points,
>            even if the same filesystem is mounted on both.)

src/builder/commands/ubuntu.rs
tailhook: You create directory `.vagga/.roots/.tmp.container.deadbee/upper` (or `x/cache` or `x/tmp_cache` or whatever) which contains cache, temporarily. Then you copy all the data into `.vagga/.cache` for persistent storage, and the original "upper" directory is left besides the container in `.vagga/.roots/container.deadbee/upper`, while it's useless right after build.

tailhook: Ah, I see it's done now in `guard.rs`

src/builder/context.rs
tailhook: Yes it's upper for overlayfs, but is not clear what it is upper for. `cache` is fine.

tailhook:
Okay. So all looks good so far.

I think we should proceed with copying files for now and maybe try to optimize (i.e. have a mountpoint that crosses both cache-dir and storage-dir) later. With the amount of file operations we have to make a container it should not make much difference.

tailhook:
I've also merged github-actions support, so please rebase on top of that, so we see tests here.

anti-social:
@tailhook Could you review it

tailhook:
`.owner(CopyOwnerPolicy::Set(..` is a bit verbose, but since used only single time looks fine.

Also `Preserve` should be removed instead of commented out.

The failing test is the flaky one, right?

Other than that it looks good to me.

anti-social:
Also updated `Lock`. It is now possible to clean up lock file.

tailhook:
Merged. Thanks!

PR #570, opened 2022-01-02
Use wait-all-successful mode for supervise flaky tests

anti-social:


tests/generic/vagga.yaml
```diff
@@ -338,42 +322,52 @@ commands:
...
+        volumes:
+          /srv: !Persistent fail-fast
         run: |
           echo hello
-          sleep 0.1
           echo world
+          while [ ! -f /srv/hanging.child ]; do
+            sleep 0.1
+          done
+          rm -f /srv/hanging.child
```
tailhook, line 360: This will work on github actions, but may fail (i.e produce unexpected success) locally and on our buildbot CI.

Can we use some flock command, which is more reliable?

tests/generic/vagga.yaml
```diff
@@ -307,6 +292,7 @@ commands:
 
   args: !Supervise
     description: Test of arg parsing
+    mode: wait-all-successful
```
tailhook, line 295: I'm not sure I understand. You use only `mode: wait-all-successful` in tests, actual we mainly run vagga in default mode.

So at a glance it looks like makes tests successful but doesn't test what was intended. Or am I missing something?

tests/generic/vagga.yaml
anti-social: Why may it fail? We always remove `.vagga` directory before running tests.

Cannot thing of how `flock` cannot be used here. We need something like java's `CountDownLatch`.

tests/generic/vagga.yaml
anti-social: Those tests are used to test other features. `wait-all-successful` just simplifies writing test cases.

Also there are tests with default mode.

tests/generic.bats
```diff
@@ -299,205 +322,223 @@ setup() {
...
-    [[ $(cat .vagga/unzip-local/root/test/4/file2.txt) = "2" ]]
-    [[ ! -d .vagga/unzip-local/root/configs/4/dir ]]
-    [[ ! -f .vagga/unzip-local/root/test/4/install.sh ]]
-    [[ $(cat .vagga/unzip-local/root/test/5/file.txt) = "Hello" ]]
-    [[ $(cat .vagga/unzip-local/root/test/5/file2.txt) = "2" ]]
-    [[ ! -d .vagga/unzip-local/root/configs/5/dir ]]
-    [[ ! -f .vagga/unzip-local/root/test/5/install.sh ]]
-    [[ ! -f $cached_file ]]
+    assert_equal $link ".roots/unzip-local.9579aef7/root"
+    assert_equal $(cat .vagga/unzip-local/root/test/1/dir/file.txt) "Hello"
```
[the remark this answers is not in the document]
anti-social: Yeah. I'll fix that.

tests/generic/vagga.yaml
tailhook: Ah, yes, sorry. Looks good.

tests/generic/vagga.yaml
tailhook: Okay. If you're sure about that, let's fix the quotes and we are good to merge.

PR #571, opened 2022-01-06
Update eatmydata; try to activate it by default

anti-social:


src/builder/commands/ubuntu.rs
```diff
@@ -675,57 +675,38 @@ impl Distro {
...
-                self.codename.as_ref().unwrap(), &self.config.arch);
-            if let Some(ref params) = eatmy {
-                if params.needs_universe {
-                    debug!("Add Universe for eat my data");
-                    self.enable_universe()?;
-                    self.add_universe(ctx)?;
+                if let Ok(()) = apt_get_install(ctx, &[params.package], &EatMyData::No) {
+                    self.eatmydata = params.find(ctx);
+                } else {
+                    warn!("Could not install {} package", params.package);
```
tailhook, line 697: Can we also emit error here?

src/builder/commands/ubuntu.rs
```diff
@@ -1226,37 +1209,53 @@ mod build {
...
+                        info!("Eatmydata activated: {}", p);
+                        EatMyData::Preload(p)
+                    }
+                )
+        }
+
+        pub fn _find(&self, ctx: &mut Context) -> Result<String, String> {
+            let emd_pkg_output = capture_command(
+                ctx,
+                &["/usr/bin/dpkg-query".to_string(), "-L".to_string(), self.package.to_string()],
```
tailhook, line 1243: Is it fast enough to run?

src/builder/commands/ubuntu.rs
anti-social: I'm afraid in future ubuntu releases eatmydata package may be renamed.

src/builder/commands/ubuntu.rs
anti-social: Fast enough:
```
$ time dpkg-query -L git

real    0m0.005s
user    0m0.000s
sys     0m0.005s
```

src/builder/commands/ubuntu.rs
anti-social: So in this way we won't break future ubuntu versions.

anti-social: Oh, you mean `error!` )))

tailhook: No I mean:
```
 match apt-get-install... {
 Err(e) => warn!("some intro {:#}", e), // << so we see actual error here
 }
```

anti-social: I think it does not add any new information (we say clearly which package cannot be installed):
```
Could not install libeatmydata1 package. The build may be slower than usual: error running <Command "/usr/bin/apt-get" "-oDir::cache::pkgc
ache=" "-oDir::cache::srcpkgcache=" "install" "-y" "libeatmydata1-asdf"; environ[5]; chroot="/vagga/root"; work-dir="/work"> exited with code 100
```

tailhook: Yes, but this adds information:
1. Whether the apt-get command could run (i.e. /usr/bin/apt-get is executable)
2. Whether all the respective volumes could be mounted
etc.

There are so many things that could go wrong beyond non-existence of the package. And if everything is okay, we won't get any message. So it's better to add more info if it fails.

anti-social:
Updated.

tailhook:
Great! Thank you.

PR #573, opened 2022-01-11
Use vagga on github CI

anti-social:
Tested cargo and vagga caches inside my fork: https://github.com/anti-social/vagga/runs/4776366943

We could add `vagga _version_hash rust-musl` to the caches keys. What do you think?

### DIFF
--- a/vagga.yaml
+++ b/vagga.yaml
@@ -11,7 +11,6 @@ containers:
     environ: &rustenv
       LD_LIBRARY_PATH: /musl/lib/rustlib/x86_64-unknown-linux-musl/lib
       PATH: /musl/bin:/usr/local/bin:/usr/bin:/bin
-      HOME: /work/target
     setup:
     - !Ubuntu focal
     - !UbuntuUniverse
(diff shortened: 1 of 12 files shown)


vagga.yaml
```diff
@@ -70,13 +70,13 @@ containers:
...
     - !BuildDeps [wget]
-    - !Install [make, curl, zip, file, git, parallel]
+    - !Install [make, curl, zip, file, git, ca-certificates, parallel]
     - !Sh |
         set -ex
         cd /tmp
-        git clone git://github.com/bats-core/bats-core
-        git clone git://github.com/bats-core/bats-support.git /bats/bats-support
-        git clone git://github.com/bats-core/bats-assert.git /bats/bats-assert
+        git clone --depth=1 https://github.com/bats-core/bats-core.git
```
tailhook, line 77: The `.git` used to be optional, isn't it?

tailhook (approved):
> We could add `vagga _version_hash rust-musl` to the caches keys. What do you think?

Yes, sounds good. This will make cache smaller (i.e. only the latest image version will be in the cache, not all versions that happened to be there until some inadvertent cache clean).

anti-social:
Also I think about making a separate cache for project's vagga commands. So when we'll update rust version test cache won't be ivalidated.

tailhook:
> Also I think about making a separate cache for project's vagga commands. So when we'll update rust version, tests cache won't be ivalidated.

Well, updating rust isn't frequent operation. So if it's not single-line fix or similar, I wouldn't bother.

anti-social:
Often see the following error on the github CI: 
```
E: Failed to fetch https://ubuntu.zerogw.com/pool/main/v/vagga/vagga_0.8.1-83-gd1f1466-1_amd64.deb  Could not handshake: The TLS connection was non-properly terminated. [IP: 148.251.14.186 443]
```

What could the reason for this?

anti-social:
@tailhook Could you review and merge if everything is ok

tailhook (approved):

tailhook:
Sure. Merged!

PR #575, opened 2022-01-24
Fix removing symlink to a directory when processing data-dirs option

anti-social:


anti-social:
@tailhook Could you review this PR

tailhook:
It's fine. But I don't understand why tests fail :(

anti-social:
It looks like `pip` bumped minimum supported python version to 3.7:
```
ERROR: This script does not work on Python 3.6 The minimum supported Python version is 3.7. Please use https://bootstrap.pypa.io/pip/3.6/get-pip.py instead.
```

tailhook:
Ah, well, this is a new failure. So previous was probably some multi-day hiccup of github :(

PR #576, opened 2022-02-04
Try to download get-pip script for specific python version, then fallback to default get-pip

anti-social:


tailhook:
Merged. Thanks!

Issue #577, opened 2022-06-08
New release? Not able to build from source.


sebastian-toepfer:
can you please provide a new version with a fix for #545?

Had tried to build it from source but too stupid to do so.

sebastian-toepfer:
sry had forget to change my sources.list

PR #578, opened 2022-06-09
bump alpines packages

sebastian-toepfer:
change version of alpine package where older version no longer be available

will fix one of my build issues from #577


sebastian-toepfer:
can somebody explain why thsi build is failed. i#m not able to understand the issue :(.

tailhook:
Basically we write tests based on the real dependencies. And sometimes those things got deprecated:
```
Composer 2.3.0 dropped support for PHP <7.2.5 and you are running 5.6.27, please upgrade PHP or use Composer 2.2 LTS via "composer self-update --2.2". Aborting.
```
So we need to keep vagga's tests updated constantly, which I'm struggling to do.

This is not your fault.

sebastian-toepfer:
@tailhook : any way how I can reproduce it? and provide a fix for it.

tailhook:
> @tailhook : any way how I can reproduce it? and provide a fix for it.

Just run tests on your machine. I.e. `vagga test`. Sometimes you may need to clean local cache first:
```
rm -rf tmp/
rm -rf .vagga
vagga test
```

sebastian-toepfer:
I'm so stupid, of course, vagga for all. Eat your own dogfood. And this one is tasty :)

@tailhook: Would this fix from ebf60eb be ok? I know is not the primary problem, but not on my machine it was an error.

sebastian-toepfer:
so the problem is, that we install an old php version with composer 2.3.x. but i do not found a way to install 2.2 :(
`src/builder/commands/composer` is the right place? and `vagga make && vagga test` should build a new version and use it to run the tests right?

